PR #158, opened 2014-08-10
Use libraries’ existing Xcode projects.

fabiopelosin:
This patch as it is looks good and I consider it acceptable to be merged.

Before doing that I'm looking for an option of the @CocoaPods/core team regarding:

- Enforce one project per spec/subspec.
- Enforce in the linter the presence of either build settings and file patterns (traditional way) or a project and target (project way).
- Also I'm not sure about the syntax:
  - `xcodeprojs` could be `project`
  - it would be cleaner to have just `project` and `target` as attributes instead of a hash. If one of the two is specified also the other must be (`project way`). This can be enforced in the linter.

--

Note, for the sake of the discussion this pull is pointing to the `lookback-pods-by-config` branch (which is expected to be merged in master soon)

### DIFF
--- a/lib/cocoapods-core/specification/consumer.rb
+++ b/lib/cocoapods-core/specification/consumer.rb
@@ -88,6 +88,11 @@ module Pod
       #
       spec_attr_accessor :xcconfig
 
+      # @return [Array<Hash>] A list of Hashes describing the subprojects to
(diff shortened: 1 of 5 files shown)


segiddins:
Closing due to lack of activity.

PR #215, opened 2015-01-11
Add the lost space preceding the attribute name in validation messages

mrackwitz:
This has caused an error after updating the gem on trunk.
```diff
- "Missing required attribute `license`."
+ "Missing required attribute`license`."
```
This minor change fixes that and causes that the generated messages will be generated equal to the prior state, seen as above in red.


segiddins:
:+1:

PR #216, opened 2015-01-13
[Analyzer] The linter will allow non-file-pattern values for `requires_arc`.

segiddins:
\c @kylef


spec/specification/linter/analyzer_spec.rb
```diff
@@ -130,6 +130,12 @@ module Pod
           results.first.message.should.include?('spec is empty')
           results.first.attribute_name.should.include?('File Patterns')
         end
+
+        it 'allows a non-string value for requires_arc' do
+          @spec.requires_arc = true
```
kylef: Wouldn't this test pass regardless of your change to `lib/cocoapods-core/specification/linter/analyzer.rb`?

PR #218, opened 2015-02-14
[Specification] Allow specifying a custom module map file

segiddins:
\c @kylef @mrackwitz 

- [x] ~~Lint to prevent module_map defined on a subspec.~~ depends on #233.
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
   [Samuel Giddins](https://github.com/segiddins)
   [#221](https://github.com/CocoaPods/Core/issues/221)
 
+* Allow specifying a custom module map file.  
+  [Samuel Giddins](https://github.com/segiddins)
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#218](https://github.com/CocoaPods/Coore/issues/218)
+
 ##### Bug Fixes
 
 * The linter will now ensure that subspecs' names do not contain whitespace.  
--- a/lib/cocoapods-core/specification/linter.rb
+++ b/lib/cocoapods-core/specification/linter.rb
@@ -210,10 +210,6 @@ module Pod
               'from default')
           end
         end
-
-        if a.empty?
-          results.add_error('authors', 'The authors are unspecified.')
-        end
       end
 
       def _validate_version(v)
--- a/spec/specification/dsl_spec.rb
+++ b/spec/specification/dsl_spec.rb
@@ -253,6 +253,11 @@ module Pod
         @spec.module_name = 'Three20'
         @spec.attributes_hash['module_name'].should == 'Three20'
       end
+
+      it 'allows to specify a custom module map file' do
+        @spec.module_map = 'module.modulemap'
+        @spec.attributes_hash['module_map'].should == 'module.modulemap'
+      end
     end
 
     #-----------------------------------------------------------------------------#
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1162,7 +1162,6 @@ module Pod
       attribute :module_map,
                 :root_only     => true
 
-
       #-----------------------------------------------------------------------#
 
       # @!group Subspecs
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1140,6 +1140,29 @@ module Pod
                 :file_patterns => true,
                 :singularize   => true
 
+      #------------------#
+
+      # @!method module_map=(module_map)
+      #
+      #   The module map file that should be used when this pod is integrated as
+      #   a framework.
+      #
+      #   ---
+      #
+      #   By default, CocoaPods creates a module map file based upon the public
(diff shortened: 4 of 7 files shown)


orta:
So this is a breaking change that would require a Specs repo update if someone pushes a pod from a beta/master to the repo, are there any other DSL attributes we should add if we're doing this?

segiddins:
Didn't we already add the module_name attribute?

-Samuel E. Giddins

On Feb 14, 2015, at 9:42 AM, Orta <notifications@github.com> wrote:

So this is a breaking change that would require a Specs repo update if someone pushes a pod from a beta/master to the repo, are there any other DSL attributes we should add if we're doing this?

orta:
Legit point, and it's being used, ignore me.

screen shot 2015-02-14 at 12 47 48 pm

segiddins:
I think this was one of the huge benefits of switching to JSON

mrackwitz:
So with subspecs, I see the following two cases:
* **Without custom modulemap**: The generated modulemap will inject the generated umbrella header, which takes care of importing all public headers, including those of subspecs. Importing the full Clang module, will give you everything. But you can also import implicitly derived submodules by using e.g. `MyPod.MySubspecHeader`.

* **With custom modulemap**: The custom modulemap will be used and injected by the build setting in the generated target. Because this is given, there is also no need to generate an umbrella header. All public headers, even those of the subspecs, need to be transitively imported or known to the modulemap by explicitly enumerating them. In consequence, the pod authors would need to make sure, that they provide a public header with precompiler-guards with conditions for header-availability for each subspec (and eventually have something like a sub-umbrella-header for each subspec), so that the resulting Clang module does include the fulll public interface, including the subspecs.

segiddins:
@mrackwitz sure. I think specifying your own module map is like 'hard mode' -- you're responsible for it doing the things it should. I don't think we'll recommend that anyone use it -- it'll just be an option for those who _need_ it. As to dealing with subspecs -- sure, that'd be nice, but (a) that would all be inside CocoaPods/CocoaPods and (b) it's 1/2 unrelated to this -- it can be done separately.

mrackwitz:
>but (a) that would all be inside CocoaPods/CocoaPods and (b) it's 1/2 unrelated to this -- it can be done separately.

We could i.e. go a whole different approach and allow customizing the generated modulemap in the DSL, so that you could add modulemap-declarations also on the subspec level.
It's a bit of *intentional* overthinking as it isn't based yet on actual issues. *Intentional* because changing the DSL back and forth is limited, so we should take care and allow an extensible design, which is compatible with existing features.
After all, this seems to be given, so from my side, we're definitively good to go for this approach. :+1:

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Master
 
+##### Enhancements
+
+* Allow specifying a custom module map file.
```
segiddins: two spaces

segiddins:
Other than the one CHANGELOG comment and the linting, this should be good to go once the work in CocoaPods is done and https://github.com/CocoaPods/Core/pull/233 has been finished.

orta:
This'll need a rebase :+1:

mrackwitz:
Done! With the change of 1375f0f, which affects how the linter is tested, I discovered a duplicated warning, introduced with #233, which I fixed right here on this branch.

lib/cocoapods-core/specification/linter.rb
```diff
@@ -210,10 +210,6 @@ def _validate_authors(a)
               'from default')
           end
         end
-
-        if a.empty?
```
segiddins: why is this removed?

spec/specification/dsl_spec.rb
```diff
@@ -253,6 +253,11 @@ module Pod
         @spec.module_name = 'Three20'
         @spec.attributes_hash['module_name'].should == 'Three20'
       end
+
+      it 'allows to specify a custom module map file' do
```
segiddins: `it 'allows specifying a custom module map file' do

lib/cocoapods-core/specification/linter.rb
mrackwitz: 238a282: The `author` attribute is specified in the DSL as `:required => true`, so it's linted automatically and we can rely on that. :)

spec/specification/dsl_spec.rb
mrackwitz: it's like all the others before and after that

mrackwitz:
This should be then ready now. Anything else left?

spec/specification/dsl_spec.rb
segiddins: well that's no reason to introduce more bad grammar :)

segiddins:
Nope, this is good! :+1:

spec/specification/dsl_spec.rb
mrackwitz: From a linguist point you may be right, but I'd prefer to keep everything consistent. So feel free to go for it and open an englishification-of-all-the-specs PR. :stuck_out_tongue_winking_eye:

Issue #213, opened 2015-01-05
Pod::StandardError - Unsupported class for YAML conversion Fixnum


ratazzi:
### Report

* What did you do?

  add pod using svn without `:revision` and run `pod install`, after `Podfile.lock` is updated, run `pod install` again, Pod::StandardError occurred.

* What did you expect to happen?

  Pod::StandardError - Unsupported class for YAML conversion Fixnum

### Stack

```
   CocoaPods : 0.35.0
        Ruby : ruby 2.1.4p247 (2014-09-24 revision 47703) [x86_64-darwin13.0]
    RubyGems : 2.4.2
        Host : Mac OS X 10.9.5 (13F34)
       Xcode : 6.1.1 (6A2008a)
         Git : git version 1.9.3 (Apple Git-50)
Ruby lib dir : /Users/ratazzi/.rvm/rubies/ruby-2.1-head/lib
Repositories : master - git://github.com/CocoaPods/Specs.git @ e8207654b2240ff02947f46dc5f1b2f435bc6da0
               master-1 - https://github.com/CocoaPods/Specs.git @ 1057e5b8711f327ba657f6c7f64b0cefdb04065f
               Yunyue - git@192.168.0.252:ratazzi/specs.git @ dc85221596e40e1b3e50347baa49d7dbc3053b0d
```

### Plugins

```

```

### Podfile

```ruby
# Uncomment this line to define a global platform for your project
platform :ios, "4.3"

source 'git://github.com/CocoaPods/Specs.git'

pod 'UMengAnalytics', '~> 2.2.1'
pod 'OpenUDID', :head
pod 'assets', :path => 'assets'
pod 'HXAppPlatformKit', :git => 'git@192.168.0.252:ratazzi/HXAppPlatformKit.git', :tag => 'v2.1.8'
pod 'AppIcon', :svn => 'file:///Volumes/Untitled/itools' # , :revision => "2" # If :revision is enabled, it is ok
```

### Podfile.lock

```ruby
PODS:
  - AppIcon (1.00000)
  - assets (1.00003)
  - HXAppPlatformKit (2.1.8)
  - OpenUDID (HEAD based on 1.0.0)
  - UMengAnalytics (2.2.1)

DEPENDENCIES:
  - AppIcon (from `file:///Volumes/Untitled/itools`)
  - assets (from `assets`)
  - HXAppPlatformKit (from `git@192.168.0.252:ratazzi/HXAppPlatformKit.git`, tag `v2.1.8`)
[...]

PR #224, opened 2015-03-08
[YAMLHelper] Fix the serialization of strings that have weird characters...

segiddins:
.../patterns

Closes https://github.com/CocoaPods/Core/issues/213.
Closes https://github.com/CocoaPods/CocoaPods/issues/2837.

\c @kylef @alloy @floere

### DIFF
--- a/lib/cocoapods-core/yaml_helper.rb
+++ b/lib/cocoapods-core/yaml_helper.rb
@@ -92,7 +92,7 @@ module Pod
       #
       def process_according_to_class(value, hash_keys_hint = nil)
         case value
-        when String     then value
+        when String     then YAML.dump(value).sub(/^---/, '').sub(/[.]{3}\s*$/, '').strip
         when Symbol     then ":#{value}"
         when TrueClass  then 'true'
         when FalseClass then 'false'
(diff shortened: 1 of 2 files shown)


kylef:
Nice, can you add a test to ensure that 1.2 as a Float is still a YAML number too?

segiddins:
Our serializer doesn't handle numbers

-Samuel E. Giddins

On Mar 8, 2015, at 5:12 AM, Kyle Fuller <notifications@github.com> wrote:

Nice, can you add a test to ensure that 1.2 as a Float is still a YAML number too?

floere:
@segiddins From the test alone it's hard to glean what the point of this is - help?

floere:
(Mostly confused by the --- replacement)

segiddins:
Ah. It uses Psych to ensure that we properly serialize strings, such as '1.2', which, if unquoted, appears as a FixNum.

floere:
Thanks, I see – we want to return only the YAMLified value, not a YAML document. Should the `.` be escaped? (`\.`)

segiddins:
> Thanks, I see – we want to return only the YAMLified value, not a YAML document.

Exactly, and the YAML module has no method that does that.

segiddins:
Let's merge this @kylef :smiling_imp:

PR #231, opened 2015-03-30
[Source] Make #search case sensistive even on case-insensitive HFS+

segiddins:
I'd love to do this in a faster way, and in a manner that doesn't rely on `Pathname.glob`'s results matching the underlying preserved case.

\c @kylef @samdmarshall

Closes https://github.com/CocoaPods/CocoaPods/issues/3024.
Closes https://github.com/CocoaPods/CocoaPods/issues/2910.

### DIFF
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -201,8 +201,7 @@ module Pod
       if query.is_a?(Dependency)
         query = query.root_name
       end
-      spec_dir = specs_dir + query
-      if Pathname.glob(spec_dir) == [spec_dir]
+      if specs_dir.children.select(&:directory?).map(&:basename).map(&:to_s).include?(query.to_s)
         set(query)
       end
     end
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -201,7 +201,8 @@ module Pod
       if query.is_a?(Dependency)
         query = query.root_name
       end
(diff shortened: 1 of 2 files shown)


lib/cocoapods-core/source.rb
```diff
@@ -201,7 +201,8 @@ def search(query)
       if query.is_a?(Dependency)
         query = query.root_name
       end
-      if (specs_dir + query).directory?
+      spec_dir = specs_dir + query
+      if Pathname.glob(spec_dir) == [spec_dir]
```
kylef: We should escape this path given to glob, since it could contain characters causing it to break. See https://github.com/CocoaPods/CocoaPods/commit/f4df7a6121e6bf4c04b44b414abe91f91b64f79a

alloy:
Maybe the [find module](http://ruby-doc.org/stdlib-2.2.0/libdoc/find/rdoc/Find.html) can be of help?

segiddins:
@alloy how so?

alloy:
Maybe it ends up being a faster way to implement the search/Pathname.glob (of which we have our own implementation iirc?), which is what you were mentioning.

lib/cocoapods-core/source.rb
segiddins: I don't think we need to, as there are no valid spec names that contain glob characters

segiddins:
This has been updated to no longer use a glob.

kylef:
:shipit:

Issue #177, opened 2014-09-29
linter is allowing whitespaces in subspecs


mmacvicar:
This commit 17aaae07c4498d68dc948c7d57b9aacd43e9a417 recently changed a validation to make the version optional in the string representation. However, it also started enforcing something that was not being checked there.

The old regexp was matching anything that ended in non-whitespaces and a version (ignoring the a any prefix that included whitespaces):
/(\S*) \((.*)\)/ 
The new one does not allow that (which is the right thing):
/\A(\S*)(?: \((.+)\))?\Z/

The problem is that the current linter checks for whitespaces in the spec name but it does not check that for the subspecs.

For example:
https://github.com/CocoaPods/Specs/blame/4a2a7d017cddfdec782da5bbdd2abee353549280/Specs/mopub-ios-sdk/2.2.0/mopub-ios-sdk.podspec.json

That podspec passes linter validations and works fine during a fresh pod install but an exception is raised in the next pod install or pod update:

[!] Invalid string representation for a specification: `mopub-ios-sdk/MoPubSDK Internal (2.2.0)`. The string representation should include the name and optionally the version of the Pod.

I would be happy to send a patch but first I wanted to verify that having "MoPubSDK Internal" as a subspec name is indeed not allowed.

PR #233, opened 2015-04-04
[Specification::DSL] Fixes around Linter and Analyzer

mrackwitz:
Includes a fix that `:root_only => true` attributes could been overwritten in subspecs.
Especially the change in a6d508f should been reviewed carefully.
/c @kylef @segiddins @alloy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,11 @@
   [Samuel Giddins](https://github.com/segiddins)
   [#221](https://github.com/CocoaPods/Core/issues/221)
 
-
 ##### Bug Fixes
+
+* The linter fails now if root attributes occur on subspec level.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#233](https://github.com/CocoaPods/Core/pull/233)
 
 * Inhibit warnings for pods that only have the `inhibit_warnings` option enabled
   on a subspec.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@
 
 ##### Bug Fixes
 
+* The linter will now ensure that subspecs' names do not contain whitespace.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [Joshua Kalpin](https://github.com/Kapin)
+  [Samuel Giddins](https://github.com/segiddins)
+  [#177](https://github.com/CocoaPods/Core/issues/177)
+  [#178](https://github.com/CocoaPods/Core/pull/178)
+  [#202](https://github.com/CocoaPods/Core/pull/202)
+  [#233](https://github.com/CocoaPods/Core/pull/233)
+
 * The linter fails now if root attributes occur on subspec level.  
   [Marius Rackwitz](https://github.com/mrackwitz)
   [#233](https://github.com/CocoaPods/Core/pull/233)
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -794,8 +794,7 @@ module Pod
       #   @param  [String] name
       #           the module name.
       #
-      root_attribute :module_name,
-                     :inherited => true
+      root_attribute :module_name
 
       #------------------#
 
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -76,8 +76,10 @@ module Pod
       #   @param  [String] name
       #           the name of the pod.
       #
-      root_attribute :name,
-                     :required => true
+      attribute :name,
+                :required => true,
+                :inherited => false,
+                :multi_platform => false
 
       #------------------#
 
--- a/lib/cocoapods-core/specification/linter.rb
+++ b/lib/cocoapods-core/specification/linter.rb
@@ -46,7 +46,7 @@ module Pod
       def lint
         @results = Results.new
         if spec
-          check_required_root_attributes
+          check_required_attributes
           validate_name
           run_root_validation_hooks
           perform_all_specs_analysis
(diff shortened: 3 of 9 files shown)


lib/cocoapods-core/specification/dsl.rb
```diff
@@ -76,8 +76,10 @@ module DSL
       #   @param  [String] name
       #           the name of the pod.
       #
-      root_attribute :name,
-                     :required => true
+      attribute :name,
```
segiddins: name should definitely be a root attribute...

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -794,8 +796,7 @@ def dependency(*args)
       #   @param  [String] name
       #           the module name.
       #
-      root_attribute :module_name,
-                     :inherited => true
+      root_attribute :module_name
```
segiddins: why the removal?

lib/cocoapods-core/specification/linter/analyzer.rb
```diff
@@ -117,12 +127,27 @@ def value_for_attribute(attribute)
           nil
         end
 
+        # Validates the occurrence of the given attribute.
```
segiddins: this comment could definitely be more explanatory

lib/cocoapods-core/specification/dsl.rb
mrackwitz: But it's also valid and required for subspecs.

lib/cocoapods-core/specification/linter/analyzer.rb
```diff
@@ -117,12 +127,27 @@ def value_for_attribute(attribute)
...
+          end
+        end
+
         # Validates the given value for the given attribute.
         #
         # @param  [Spec::DSL::Attribute] attribute
         #         The attribute.
         #
-        # @param  [Spec::DSL::Attribute] value
+        # @param  [mixed] value
```
segiddins: `Object`

lib/cocoapods-core/specification/dsl.rb
mrackwitz: Because root_attributes can't be inherited.

lib/cocoapods-core/specification/dsl.rb
segiddins: ah but you shouldn't be able to set the subspecs name directly, it's set via `spec.subspec 'ss_name` do |ss|` bit

mrackwitz: Agree on that, but if it's defined as `root_attribute`, it will be never validated for subspecs and the behavior, when the value is set, is undefined. It would be an exception among other root attributes, because it may occur on subspec level. I think being to overwrite it in the subspec block in the DSL is the lesser of the two evils.

lib/cocoapods-core/specification/linter/analyzer.rb
mrackwitz: :+1:

lib/cocoapods-core/specification/dsl.rb
segiddins: @mrackwitz in that case, we must validate that
```ruby
if spec.parent
  spec.name.start_with?("#{spec.parent.name}/")
end
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Master
 
+##### Enhancements
+
+* The linter fails now if root attributes occur on subspec level.
```
segiddins: two spaces

lib/cocoapods-core/specification/linter.rb
```diff
@@ -46,7 +46,8 @@ def initialize(spec_or_path)
       def lint
         @results = Results.new
         if spec
-          check_required_root_attributes
+          check_required_attributes
+          validate_name
```
segiddins: couldn't this just be `_validate_name`, and then wouldn't it be picked up automatically?

mrackwitz: Not if name would be not a root attribute anymore.

segiddins:
This needs more testing around setting names in subspecs

lib/cocoapods-core/specification/dsl.rb
mrackwitz: @segiddins: Where would you add that check?

segiddins: linter?

mrackwitz: That's not required to lint `Specification`, because the names are not flattened. Subspecs store only the last name element in their `attributes_hash[:name]`.

segiddins: :+1: in that case, need to validate that no `/` gets into `attributes_hash[:name]`

mrackwitz: Yes, indeed. :+1:

mrackwitz:
@segiddins: Added the linter validation for subspec names. Anything else, you see?

spec/specification/linter_spec.rb
```diff
@@ -145,6 +150,11 @@ def result_should_include(*values)
         result_should_include('name', 'whitespace')
       end
 
+      it 'fails a specification whose name contains a backslash' do
+        @spec.stubs(:name).returns('BananaKit/BananaFruit')
```
kylef: The test name says backslash but here we are testing a forward slash.

segiddins:
@mrackwitz while you're at it, please make sure https://github.com/CocoaPods/Core/issues/177 and https://github.com/CocoaPods/Core/pull/178 are addressed?

lib/cocoapods-core/specification/linter.rb
```diff
@@ -180,6 +178,11 @@ def _validate_name(_n)
                            'name of the file.')
           end
 
+          if spec.name =~ /\//
+            results.add_error('name', 'The name of a spec should not contain ' \
+                           'a slash.')
+          end
+
           if spec.root.name =~ /\s/
```
segiddins: if `spec.name`

mrackwitz:
@segiddins: Addressed the issues.

CHANGELOG.md
```diff
@@ -9,6 +9,18 @@
   [Samuel Giddins](https://github.com/segiddins)
   [#221](https://github.com/CocoaPods/Core/issues/221)
 
+* The linter will now ensure that subspecs' names do not contain whitespace.
```
kylef: IMO, this is a bug fix not an enhancement. (same goes for below entry)

mrackwitz: Overall it will be likely rather perceived as bug fixes.

Issue #221, opened 2015-02-26
List all desired subspecs on one line in Podfile


mickeyreiss:
Similar to #150, I am also hitting a point where some sugar on top of subspecs would be useful.

In my case, I'd like to have a single Pod with ~5 optional subcomponents. Each user of the pod is likely to choose a different subset; it's nearly impossible to determine what the right set of defaults should be. By choosing a subset for people, we likely introduce bloat into typical integrators' apps.

In this type of setup, where subspecs are closer to build variants than separate components, I propose a syntax that allows the user to specify all relevant subspecs on one line in their `Podfile`:

```ruby
pod "CuttingTools", :subspecs => ["Scissors", "Machette", "Penknife", "KnifeBlock"]
```

as an alternative to

```ruby
pod "CuttingTools/Scissors"
pod "CuttingTools/Machette"
pod "CuttingTools/Penknife"
pod "CuttingTools/KnifeBlock"
```

You'll note in this example that `KnifeBlock` has a hierarchical relationship with the rest of the subspecs, in that it holds whichever knives happen to be available; at the same time, if KnifeBlock pulls in every tool (subspec) available as a dependency, many users will end up with extra code provided by the unused katana, butter-knife, etc. (i.e. undesired subspecs for that application).

Has something along these lines been considered in the past?

segiddins:
We'd actually rather encourage people to not use subspecs :)

mickeyreiss:
What would you suggest as an alternative for situations like this?

segiddins:
Runtime customization?

-Samuel E. Giddins

On Feb 25, 2015, at 8:25 PM, Mickey Reiss <notifications@github.com> wrote:

What would you suggest as an alternative for situations like this?

kylef:
Since this is Ruby and it's a full language, a short term solution for this that you can currently use with CocoaPods:

```ruby
%w(Scissors Machette).each{ |s| pod "CuttingTools/#{s}" }
```

---

Perhaps the proposed syntax, or a similar syntax makes sense to simplify this syntax. @orta any input?

segiddins:
@orta do we want to add this to the DSL?

orta:
Oh wow, I have never even thought of this. It'd be lovely IMO.

PR #234, opened 2015-04-05
[TargetDefinition] Allow specifying multiple subspec dependencies inline

segiddins:
Closes https://github.com/CocoaPods/Core/issues/221.

\c @kylef @orta @alloy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ##### Enhancements
 
 * Allow specifying multiple subspec pod dependencies inline in the Podfile,
-  via `pod 'Pod', subspecs: ['Subspec1', 'Subspec2']`.  
+  via `pod 'Pod', :subspecs => ['Subspec1', 'Subspec2']`.  
   [Samuel Giddins](https://github.com/segiddins)
   [#221](https://github.com/CocoaPods/Core/issues/221)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Master
 
+##### Enhancements
+
+* Allow specifying multiple subspec pod dependencies inline in the Podfile,
+  via `pod 'Pod', subspecs: ['Subspec1', 'Subspec2']`.  
+  [Samuel Giddins](https://github.com/segiddins)
+  [#221](https://github.com/CocoaPods/Core/issues/221)
+
+
 ##### Bug Fixes
 
 * Inhibit warnings for pods that only have the `inhibit_warnings` option enabled
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -468,7 +468,7 @@ module Pod
       # @return [void]
       #
       def store_pod(name, *requirements)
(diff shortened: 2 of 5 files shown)


orta:
I am really glad about this, ARAnalytics ( and my upcoming NSURLRequests pod ) will be much nicer.

segiddins:
I think we might need to document this in the `pod` method in the DSL as well, but I just threw this together.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Master
 
+##### Enhancements
+
+* Allow specifying multiple subspec pod dependencies inline in the Podfile,
+  via `pod 'Pod', subspecs: ['Subspec1', 'Subspec2']`.
```
kylef: I'd rather we use the hash rocket syntax in this CHANGELOG entry, we should stick to a consistent way of doing things in guides/examples etc as to not confuse our users who are not familiar with Ruby.

kylef:
Nice work :+1:, looks great.

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -468,6 +468,7 @@ def set_platform(name, target = nil)
       # @return [void]
       #
       def store_pod(name, *requirements)
+        parse_subspecs(name, requirements) # This parse method must be called first
```
alloy: Is that comment really needed? Won’t the tests indicate that this is needed?

alloy:
Nice :+1:

lib/cocoapods-core/podfile/target_definition.rb
segiddins: It's not 100% needed, but I feel it's better to write it down than make someone chase thru the tests next time around. The tests describe the desired end effect, the comment addresses the _why_.

spec/podfile/dsl_spec.rb
```diff
@@ -44,6 +44,33 @@ module Pod
         target.pod_whitelisted_for_configuration?('PonyDebugger', 'Debug').should.be.false
       end
 
+      it 'allows specifying multiple additional subspecs' do
+        podfile = Podfile.new do
+          pod 'RestKit', '~> 0.24.0', :subspecs => %w(CoreData Networking), :configurations => %w(Release)
```
kylef: With this implementation, this example shows that the root spec is also included.

```ruby
pod 'RestKit', :subspecs => %w(CoreData Networking)
```

Expands into:

- RestKit
- RestKit/CoreData
- RestKit/Networking

I'm not sure if it makes sense to include the root spec when using `:subspecs` since by default, the root spec *should* make the full library available unless their is alternative incompatible subspecs.

This means that this actually have no benefit for pods such as ARAnalytics since:

```ruby
pod 'ARAnaytics', :subspecs => ['Crashlytics', 'Mixpanel']
```

Would effectively do nothing since all the default_subspecs from ARAnalytics are included (which is all of them if i'm not mistaken).

I've created a PR #236 to change it's behaviour.

segiddins:
Let's :ship: ?

kylef:
:ship:-ed

PR #235, opened 2015-04-06
[Podfile DSL] Document how subspecs can be included

kylef:


lib/cocoapods-core/podfile/dsl.rb
```diff
@@ -112,6 +112,19 @@ module DSL
       #
       # ------
       #
+      # ### Subspecs
+      #
+      # By default, all of a Pod's default subspecs are installed, unless
```
segiddins: not true, only the default subspecs

lib/cocoapods-core/podfile/dsl.rb
```diff
@@ -112,6 +112,19 @@ module DSL
...
+      # you specifically specify a subspec as follows:
+      #
+      #     pod 'QueryKit/Attribute'
+      #
+      # You may specify a collection of subspecs to be installed as follows:
+      #
+      #     pod 'QueryKit', :subspecs => ['Attribute', 'QuerySet']
```
segiddins: should this mention that it also means installing the root spec?

lib/cocoapods-core/podfile/dsl.rb
orta: Wording is a bit strange in general ( default used twice )

kylef: @segiddins That's what I said "default subspecs"?

segiddins: ah must've misread it then

kylef:
I've addressed the feedback. I'm still not completely happy with this, would appreciate any suggestions.

lib/cocoapods-core/podfile/dsl.rb
```diff
@@ -112,6 +112,23 @@ module DSL
       #
       # ------
       #
+      # ### Subspecs
+      #
+      # When installing a Pod via it's name, it will install all of the
+      # default subspecs defined in the Pod.
```
segiddins: defined in the podspec?

lib/cocoapods-core/podfile/dsl.rb
```diff
@@ -112,6 +112,23 @@ module DSL
...
+      #
+      # You may install a specific subspec using the following:
+      #
+      #     pod 'QueryKit/Attribute'
+      #
+      # You may specify a collection of subspecs to be installed as follows:
+      #
+      #     pod 'QueryKit', :subspecs => ['Attribute', 'QuerySet']
+      #
+      # *NOTE*: **Installing a subspec also installs the root spec for the pod.**
```
segiddins: That's only true if you use the `podspecs: ` shortcut, iirc

kylef: -> #236

segiddins: this is now no longer true

PR #236, opened 2015-04-06
[TargetDefinition] Don't include current dependency when using subspecs

kylef:
See https://github.com/CocoaPods/Core/pull/234/files#r27783874

### DIFF
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -468,7 +468,7 @@ module Pod
       # @return [void]
       #
       def store_pod(name, *requirements)
-        parse_subspecs(name, requirements) # This parse method must be called first
+        return if parse_subspecs(name, requirements) # This parse method must be called first
         parse_inhibit_warnings(name, requirements)
         parse_configuration_whitelist(name, requirements)
 
@@ -753,17 +753,18 @@ module Pod
       #         If :subspecs is the only key in the hash, the hash
       #         should be destroyed because it confuses Gem::Dependency.
       #
-      # @return [void]
+      # @return [Boolean] Whether new subspecs were added
       #
       def parse_subspecs(name, requirements)
         options = requirements.last
-        return requirements unless options.is_a?(Hash)
+        return false unless options.is_a?(Hash)
 
         subspecs = options.delete(:subspecs)
         subspecs.each do |ss|
           store_pod("#{name}/#{ss}", *requirements.dup)
         end if subspecs
         requirements.pop if options.empty?
+        !subspecs.nil?
       end
 
       #-----------------------------------------------------------------------#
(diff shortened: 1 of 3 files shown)


segiddins:
\c @orta @alloy

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -753,17 +753,18 @@ def parse_configuration_whitelist(name, requirements)
       #         If :subspecs is the only key in the hash, the hash
       #         should be destroyed because it confuses Gem::Dependency.
       #
-      # @return [void]
+      # @return [TrueClass,FalseClass] Returns true if we added new subspecs
```
segiddins: `Boolean`

segiddins: `Whether new subspecs were added`

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -753,17 +753,18 @@ def parse_configuration_whitelist(name, requirements)
...
         options = requirements.last
-        return requirements unless options.is_a?(Hash)
+        return false unless options.is_a?(Hash)
 
         subspecs = options.delete(:subspecs)
         subspecs.each do |ss|
           store_pod("#{name}/#{ss}", *requirements.dup)
         end if subspecs
         requirements.pop if options.empty?
+        !subspecs.nil?
```
segiddins: can just return `subspecs here`

kylef: subspecs is an array and not a boolean like this method is expected to return.

segiddins: Any object can be treated like a Boolean in Ruby. Returning the object is the same as returning whether the object is not nil, it's just less code. 

-Samuel E. Giddins

On Apr 6, 2015, at 5:41 AM, Kyle Fuller <notifications@github.com> wrote:

In lib/cocoapods-core/podfile/target_definition.rb:

>  
>          subspecs = options.delete(:subspecs)
>          subspecs.each do |ss|
>            store_pod("#{name}/#{ss}", *requirements.dup)
>          end if subspecs
>          requirements.pop if options.empty?
> +        !subspecs.nil?
subspecs is an array and not a boolean like this method is expected to return.

kylef: Not exactly, if it's documented to return a Boolean it should 100% return a boolean.

```ruby
> [].is_a?(TrueClass)
=> false
> [] == true
=> false
```

segiddins: That's not true by convention, if it documents it returns a Boolean, it just means you can only treat the return value as a Boolean. 

-Samuel E. Giddins

On Apr 6, 2015, at 8:03 AM, Kyle Fuller <notifications@github.com> wrote:

In lib/cocoapods-core/podfile/target_definition.rb:

>  
>          subspecs = options.delete(:subspecs)
>          subspecs.each do |ss|
>            store_pod("#{name}/#{ss}", *requirements.dup)
>          end if subspecs
>          requirements.pop if options.empty?
> +        !subspecs.nil?
Not exactly, if it's documented to return a Boolean it should 100% return a boolean.

> [].is_a?(TrueClass)
=> false
> [] == true
=> false

PR #242, opened 2015-05-17
[WIP] Split xcconfig to pod_target_xcconfig and user_target_xcconfig

mrackwitz:
* [x] Implement the deprecation logic.

Belongs to CocoaPods/CocoaPods#3573.

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -717,18 +717,18 @@ module Pod
 
       #------------------#
 
-      # @!method xcconfig=(value)
+      # @!method pod_target_xcconfig=(value)
       #
-      #   Any flag to add to the final xcconfig file.
+      #   Any flag to add to the final __private__ pod target xcconfig file.
       #
       #   @example
       #
-      #     spec.xcconfig = { 'OTHER_LDFLAGS' => '-lObjC' }
+      #     spec.pod_target_xcconfig = { 'OTHER_LDFLAGS' => '-lObjC' }
       #
       #   @param  [Hash{String => String}] value
-      #           A representing an xcconfig.
+      #           Key-value pairs representing build settings.
       #
-      attribute :xcconfig,
+      attribute :pod_target_xcconfig,
                 :container => Hash,
                 :inherited => true
 
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -732,6 +732,41 @@ module Pod
                 :container => Hash,
                 :inherited => true
 
+      # @!method user_target_xcconfig=(value)
+      #
+      #   Any flag to add to the final aggregate target xcconfig file, which
+      #   propagates to non-overridden build settings to the integrated user
+      #   targets.
+      #
+      #   ---
+      #
+      #   This attribute is __not recommended__ as Pods should not pollute the
+      #   build settings of the user project and this can cause conflicts.
+      #
+      #   Multiple definitions for build settings, which take multiple values,
+      #   will be merged. The user is warned on conflicting definitions for
+      #   custom build settings and build settings, which take only one value.
+      #
+      #   Typically clang compiler flags or precompiler macro definitions go
+      #   in here, if they are required when importing the pod in the user
+      #   target. Note that, this influences only the compiler view of the
+      #   public interface of your pod, but also all other integrated pods
+      #   alongside to yours. You should always prefer [`pod_target_xcconfig`](
+      #   http://guides.cocoapods.org/syntax/podspec.html#pod_target_xcconfig),
+      #   which can contain the same settings, but only influence the
+      #   toolchain, when compiling your pod target.
+      #
+      #   @example
+      #
+      #     spec.user_target_xcconfig = { 'MY_SUBSPEC' => 'YES' }
+      #
+      #   @param  [Hash{String => String}] value
+      #           Key-value pairs representing build settings.
+      #
+      attribute :user_target_xcconfig,
+                :container => Hash,
+                :inherited => true
+
       #------------------#
 
       # @!method prefix_header_contents=(content)
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -734,27 +734,27 @@ module Pod
 
       # @!method user_target_xcconfig=(value)
       #
-      #   Any flag to add to the final aggregate target xcconfig file, which
-      #   propagates to non-overridden build settings to the integrated user
-      #   targets.
+      #   Specifies flags to add to the final aggregate target xcconfig file,
+      #   which propagates to non-overridden and inheriting build settings to
+      #   the integrated user targets.
       #
       #   ---
       #
       #   This attribute is __not recommended__ as Pods should not pollute the
       #   build settings of the user project and this can cause conflicts.
       #
-      #   Multiple definitions for build settings, which take multiple values,
+      #   Multiple definitions for build settings that take multiple values
       #   will be merged. The user is warned on conflicting definitions for
-      #   custom build settings and build settings, which take only one value.
+      #   custom build settings and build settings that take only one value.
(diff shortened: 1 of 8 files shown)


segiddins:
Can we think of a more concise name for those attributes? Think along the lines of public/private.

segiddins:
What's the reason for keeping the user target xcconfig option at all?

lib/cocoapods-core/specification/dsl/deprecations.rb
```diff
@@ -9,6 +9,13 @@ def preferred_dependency=(name)
           CoreUI.warn "[#{self}] `preferred_dependency` has been renamed "\
             'to `default_subspecs`.'
         end
+
+        def xcconfig=(name)
```
segiddins: For backwards compatibility, won't this need to set both?

mrackwitz:
> Can we think of a more concise name for those attributes? Think along the lines of public/private. 

I was thinking about that, but in my opinion, this is just masking the meaning as you would usually expect that something public is also visible/accessible from the private area. But those here shouldn't be inherited like it happens for `PublicPodXCConfig`. Those names used here are very explicit about the effect range.

>What's the reason for keeping the user target xcconfig option at all?

I think this discussion should go in CocoaPods/CocoaPods#3465.

lib/cocoapods-core/specification/dsl/deprecations.rb
mrackwitz: Good point! :+1:

mrackwitz:
@orta: Can you review [the documentation of the new attribute](https://github.com/CocoaPods/Core/pull/242/files#diff-0c2cb09f968f1dd4c986d5e00ac9a4aeR737)?

lib/cocoapods-core/specification/dsl/deprecations.rb
```diff
@@ -9,6 +9,13 @@ def preferred_dependency=(name)
           CoreUI.warn "[#{self}] `preferred_dependency` has been renamed "\
             'to `default_subspecs`.'
         end
+
+        def xcconfig=(value)
+          self.pod_target_xcconfig = value
+          self.user_target_xcconfig = value
+          CoreUI.warn "[#{self}] `xcconfig` has been splitted to "\
```
segiddins: `has been split between`

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -717,18 +717,53 @@ def dependency(*args)
...
+      #           Key-value pairs representing build settings.
       #
-      attribute :xcconfig,
+      attribute :pod_target_xcconfig,
+                :container => Hash,
+                :inherited => true
+
+      # @!method user_target_xcconfig=(value)
+      #
+      #   Any flag to add to the final aggregate target xcconfig file, which
```
orta: > Any flag to add

Specifies flags to add to

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -717,18 +717,53 @@ def dependency(*args)
...
+      #   propagates to non-overridden build settings to the integrated user
+      #   targets.
+      #
+      #   ---
+      #
+      #   This attribute is __not recommended__ as Pods should not pollute the
+      #   build settings of the user project and this can cause conflicts.
+      #
+      #   Multiple definitions for build settings, which take multiple values,
```
orta: > for build settings, which take multiple values,

for build settings that take multiple values 

( remove the comma too )

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -717,18 +717,53 @@ def dependency(*args)
...
+      #   will be merged. The user is warned on conflicting definitions for
+      #   custom build settings and build settings, which take only one value.
```
orta: > settings, which

settings that

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -717,18 +717,53 @@ def dependency(*args)
...
+      #
+      #   Typically clang compiler flags or precompiler macro definitions go
+      #   in here, if they are required when importing the pod in the user
```
orta: > in here, if they

in here if they

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -717,18 +717,53 @@ def dependency(*args)
...
+      #   target. Note that, this influences only the compiler view of the
```
orta: > this influences only the compiler view

this influences not only the compiler view

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -717,18 +717,53 @@ def dependency(*args)
...
+      #   public interface of your pod, but also all other integrated pods
+      #   alongside to yours. You should always prefer [`pod_target_xcconfig`](
+      #   http://guides.cocoapods.org/syntax/podspec.html#pod_target_xcconfig),
+      #   which can contain the same settings, but only influence the
+      #   toolchain, when compiling your pod target.
```
orta: > toolchain, when compiling

toolchain when compiling

orta:
done'd - thanks for the shout out

mrackwitz:
@orta: Thank you for all the feedback. It's addressed now.

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -717,18 +717,53 @@ def dependency(*args)
...
+      attribute :pod_target_xcconfig,
+                :container => Hash,
+                :inherited => true
+
+      # @!method user_target_xcconfig=(value)
+      #
+      #   Specifies flags to add to the final aggregate target xcconfig file,
+      #   which propagates to non-overridden and inheriting build settings to
+      #   the integrated user targets.
+      #
```
orta: > Specifies flags to add to the final aggregate target xcconfig file. These can be overridden by a library consumer. During integration CocoaPods will warn the consumer if the build setting does not contain `$(inherited)` if they have been changed.

mrackwitz:
I just realized that the deprecations only work for the Ruby DSL. JSON podspecs remain unaffected from that, which would mean that we would need to re-write all specs and instantly loose backwards-compatibility or we introduce deprecation logic into the `Specification` model. I tend to prefer the latter. Any thoughts on that? /c @kylef

orta:
+1 on deprecation logic

segiddins:
@mrackwitz :+1: on deprecation logic when reading from the `Specification`. Please also add a lint step that validates that `xcconfig` and the public/private ones are used mutually exclusively?

mrackwitz:
@segiddins: Good idea!

lib/cocoapods-core/specification/dsl/deprecations.rb
```diff
@@ -9,6 +9,13 @@ def preferred_dependency=(name)
           CoreUI.warn "[#{self}] `preferred_dependency` has been renamed "\
             'to `default_subspecs`.'
         end
+
+        def xcconfig=(value)
+          self.pod_target_xcconfig = value
+          self.user_target_xcconfig = value
+          CoreUI.warn "[#{self}] `xcconfig` has been split to "\
```
segiddins: `into`

segiddins:
@mrackwitz please review my latest commit

segiddins:
@mrackwitz we good to merge this?

mrackwitz:
@segiddins: Let me merge the PR on Core, rebase this, update the bundle and rebuild the integration specs, then we are good to go.

segiddins:
:+1: I'll handle that right now :)

kylef:
@segiddins are you still handling this?

https://github.com/CocoaPods/CocoaPods/pull/3573 hasn't been merged yet.

segiddins:
@kylef no, @mrackwitz told me he wanted to handle this

mrackwitz:
I did. https://github.com/CocoaPods/CocoaPods/pull/3573 should be ready now.
But I didn't realized earlier that we need to get that merged before anything else to maintain a releasable `master` on the CocoaPods repo.

PR #254, opened 2015-07-27
[YAMLHelper] Allow serializing objects of arbitrary classes

segiddins:
Closes https://github.com/CocoaPods/CocoaPods/issues/3907.

\c @alloy


segiddins:
Should be green now

segiddins:
I'm merging this unless there are any objections.

alloy:
As long as deserializing arbitrary classes is not enabled it should be good.

Btw, I'm on vacation so won't see these in time ;)

PR #256, opened 2015-07-27
Account for missing numeric segments while comparing Pod Versions

gabro:
As discussed in https://github.com/CocoaPods/cocoapods.org/issues/185#issuecomment-123417647

Long story short, now

```
3.0.0-alpha.1 < 3.0-beta.9
```

I haven't been too creative: I simply handled major, minor and patch explicitly then just used the same comparison algorithm as `Gem::Version` for the remaining segments.

@floere I tried to write some tests for the "funky" versions you pointed me at (http://search.cocoapods.org/api/v1/pods.facets.json?include=version&only=version&counts=false) but apparently `Version.new("20aa2b2b15b6f8db350ec07b6041e4951bb255d0")` is not a valid `Pod::Version` (the initializer fails)

Same goes for other version strings such as `"beta.9"`, `"2-beta"` and similar.

### DIFF
--- a/lib/cocoapods-core/version.rb
+++ b/lib/cocoapods-core/version.rb
@@ -150,8 +150,8 @@ module Pod
     #
     # @note   Attempts to compare something that's not a {Version} return nil
     #
-    def <=> other
-      return unless Pod::Version === other
+    def <=>(other)
+      return unless other.is_a?(Pod::Version)
       return 0 if @version == other.version
 
       if major != other.major
@@ -180,13 +180,13 @@ module Pod
         i += 1
 
         next      if lhs == rhs
-        return -1 if String  === lhs && Numeric === rhs
-        return  1 if Numeric === lhs && String  === rhs
+        return -1 if lhs.is_a?(String) && rhs.is_a?(Numeric)
+        return  1 if lhs.is_a?(Numeric) && rhs.is_a?(String)
 
         return lhs <=> rhs
       end
 
-        return version <=> other.version
+      version <=> other.version
     end
 
     private
--- a/lib/cocoapods-core/version.rb
+++ b/lib/cocoapods-core/version.rb
@@ -138,6 +138,51 @@ module Pod
     #
     def patch
       numeric_segments[2].to_i
+  end
+
+  ##
+  # Compares this version with +other+ returning -1, 0, or 1 if the
+  # other version is larger, the same, or smaller than this
+  # one. Attempts to compare to something that's not a
+  # <tt>Pod::Version</tt> return +nil+.
+
+  def <=> other
+    return unless Pod::Version === other
+    return 0 if @version == other.version
+
+    if major != other.major
+      return major <=> other.major
+    end
+
+    if minor != other.minor
+      return minor <=> other.minor
+    end
+
+    if patch != other.patch
+      return patch <=> other.patch
+    end
+
+    lhsegments = segments.drop_while { |s| s.is_a?(Numeric) }
+    rhsegments = other.segments.drop_while { |s| s.is_a?(Numeric) }
+
+    lhsize = lhsegments.size
+    rhsize = rhsegments.size
+    limit  = (lhsize > rhsize ? lhsize : rhsize) - 1
+
+    i = 0
+
+    while i <= limit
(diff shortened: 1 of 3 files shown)


lib/cocoapods-core/version.rb
```diff
@@ -138,6 +138,51 @@ def minor
     #
     def patch
       numeric_segments[2].to_i
+  end
```
segiddins: alignment here is off

lib/cocoapods-core/version.rb
```diff
@@ -138,6 +138,51 @@ def minor
...
+
+  ##
```
segiddins: we generally use [YARD](http://yardoc.org) to document APIs

segiddins:
This generally looks good, but please take a look at the spec failures? https://travis-ci.org/CocoaPods/Core/jobs/72843992#L1216

lib/cocoapods-core/version.rb
gabro: good catch, thanks

lib/cocoapods-core/version.rb
gabro: I admittedly copy-pasted it from `Gem::Version`, I'll update the doc style

floere:
@gabro Would it be possible to use a `super` call to avoid code duplication? Apologies if it's obviously not possible – I only am able to browse via mobile phone.

gabro:
@floere I though about it but I didn't come up with an elegant solution.

`<=>` takes an `other` version as input parameter, and compares it with `self`. If `self` has a missing patch number, we would be back to square 1.

lib/cocoapods-core/version.rb
```diff
@@ -140,6 +140,55 @@ def patch
...
+    # @param  [Version] other
+    #         The other version to compare.
+    #
+    # @return [Fixnum] -1, 0, or +1 depending on whether the receiver is less
+    #         than, equal to, or greater than other.
+    #
+    # @note   Attempts to compare something that's not a {Version} return nil
+    #
+    def <=> other
+      return unless Pod::Version === other
```
floere: Should this be `Gem::Version`?

floere:
@gabro Thanks! I see the issue – the method is too large in `Gem::Version` to cleanly code around it by using `super`.

lib/cocoapods-core/version.rb
gabro: Possibly. Do we want to allow comparison between `Gem` and `Pod` versions?

floere:
Good question. In a world where only `Pod::Version`s are used in Core, we probably don't.

I was possibly thinking of Liskov when asking this.

@segiddins @gabro In any case, with succeeding specs I am happy to pull :)

gabro:
Ok, this is trickier than I thought and somewhat related to the `Pod::Version` vs `Gem::Version` matter.
We would want to always use `Pod::Version` in comparisons, but this is not always true.

The reason specs are failing is that `Pod::Requirement` inherits `initialize` from `Gem::Requirement` and uses the `parse` from the superclass.
This results in versions that are `Gem::Version` instead of the expected `Pod::Version`, so the comparison fails.

I should have fixed it in my latest commit, although it's not pretty (code duplication).

segiddins:
Looks from the failing specs like you might need to override `none?` as well?

gabro:
Yes, I was looking into it right now.

gabro:
Ok, almost there.

There were a few style issues: I fixed most of them, but two of them I chose to ignore:

- constants naming convention for `DefaultRequirement` (since it's an override and it's used by the super implementation)

- perceived complexity of the `<=>` method. I'm not super happy with that piece of code and there's clearly room for improvement, but 90% of it is already working in production, so I'm very reluctant to refactor it.

segiddins:
thanks so much, @gabro! I'm going to make a few tweaks and add a few more tests, and then merge! Great work!

gabro:
Thanks to you guys for the great work! I'm glad I could give a little contribution!

segiddins:
Manually merged via https://github.com/CocoaPods/Core/commit/74f8fdcfd840a332fa8aee46112e26ea6a6090c8.

PR #262, opened 2015-08-27
[Podfile] Add a checksum property

segiddins:


segiddins:
\c @alloy because this is touching the YAMLHelper (again...)

alloy:
And then what is the checksum used for?

orta:
When adding or removing something like "inhibit warnings" CP cannot tell that there has been a change

segiddins:
It doesn't appear to be stable across ruby versions :(

alloy:
@orta Gotcha.

@segiddins Le sigh…

segiddins:
I guess I go the defined in file route :/

segiddins:
Please review

lib/cocoapods-core/yaml_helper.rb
```diff
@@ -109,9 +109,21 @@ def process_according_to_class(value, hash_keys_hint = nil)
...
+          processed = process_according_to_class(array_value)
+          case array_value
+          when Array, Hash
+            if array_value.size > 1
+              processed = processed.gsub(/^.*/).to_a
+              head = processed.shift
+              processed.map { |s| "  #{s}" }.prepend(head).join("\n")
+            else
+              processed
+            end
```
alloy: It’s not really clear to me, just from the code, what this is doing exactly. Does the 1 test really cover all the bases?

alloy:
Thinking about this some more, does this mean that adding e.g. a comment to the Podfile would trigger a need to re-install?

segiddins:
@alloy well this alone will have no effect, but since I can't think of an (easy) stable way to checksum the logical contents of the podfile (or podspec), I can't think of an alternative?

lib/cocoapods-core/yaml_helper.rb
segiddins: yeah -- it's basically just about making sure nested collections get properly indented, otherwise their yaml is invalid

PR #265, opened 2015-09-23
Changes search index format to fasten `pod search --full` command.

manuyavuz:
Detailed discussion can be found here: CocoaPods/cocoapods-search#8

- Previous implementation was having the drawback of traversing and performing `gsub` on all index strings for each pod specification.
- New implementation stores words as keys and list of pods containing corresponding word inside their specification as the values for corresponding hash keys. By this way, sources manager can perform a faster search operation.

### DIFF
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -288,8 +288,8 @@ module Pod
     def update(show_output = false)
       changed_spec_paths = []
       Dir.chdir(repo) do
-        prev_commit_hash = (`git rev-parse HEAD`).strip
-        output = `git pull --ff-only 2>&1`
+        prev_commit_hash = git_commit_hash
+        output = update_git_repo
         CoreUI.puts output if show_output
         unless $?.success?
           CoreUI.warn 'CocoaPods was not able to update the ' \
@@ -297,7 +297,7 @@ module Pod
                   'and persists you can inspect it running ' \
                   '`pod repo update --verbose`'
         end
-        changed_spec_paths = (`git diff --name-only #{prev_commit_hash}..HEAD`).strip.split("\n")
+        changed_spec_paths = diff_until_commit_hash(prev_commit_hash)
       end
       changed_spec_paths
     end
@@ -367,6 +367,18 @@ module Pod
           repo
         end
       end
+    end
+
+    def git_commit_hash
+      (`git rev-parse HEAD` || '').strip
+    end
+
+    def update_git_repo
+      `git pull --ff-only 2>&1`
+    end
+
+    def diff_until_commit_hash(commit_hash)
+      (`git diff --name-only #{commit_hash}..HEAD` || '').strip.split("\n")
     end
 
     #-------------------------------------------------------------------------#
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -93,6 +93,22 @@ module Pod
       end.sort
     end
 
+    # Returns pod names for given array of specification paths.
+    #
+    # @param  [Array<String>] spec_paths
+    #         Array of file path names for specifications. Path strings should be relative to the source path.
+    #
+    # @return [Array<String>] the list of the name of Pods corresponding to specification paths.
+    #
+    def pods_for_specification_paths(spec_paths)
+      spec_paths.map do |path|
+        absolute_path = repo + path
+        relative_path = absolute_path.relative_path_from(specs_dir)
+        # The first file name returned by 'each_filename' is the pod name
+        relative_path.each_filename.first
+      end
+    end
+
     # @return [Array<Version>] all the available versions for the Pod, sorted
     #         from highest to lowest.
     #
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -279,19 +279,27 @@ module Pod
     #-------------------------------------------------------------------------#
 
     # Updates the local clone of the source repo.
-    # @return  [Array] output and changed_spec_paths
-    #         First value of the array is `git pull` output. Second value is an array of changed spec paths.
     #
-    def update
-      output = nil
-      changed_spec_paths = nil
+    # @param  [Bool] show_output
+    #
+    # @return  [Array<String>] changed_spec_paths
+    #          Returns the list of changed spec paths.
+    #
+    def update(show_output = false)
+      changed_spec_paths = []
       Dir.chdir(repo) do
         prev_commit_hash = (`git rev-parse HEAD`).strip
         output = `git pull --ff-only 2>&1`
-        raise Informative, output unless $?.success?
+        CoreUI.puts output if show_output
+        unless $?.success?
+          CoreUI.warn 'CocoaPods was not able to update the ' \
+                  "`#{name}` repo. If this is an unexpected issue " \
+                  'and persists you can inspect it running ' \
+                  '`pod repo update --verbose`'
+        end
         changed_spec_paths = (`git diff --name-only #{prev_commit_hash}..HEAD`).strip.split("\n")
       end
-      [output, changed_spec_paths]
+      changed_spec_paths
     end
 
     public
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -275,6 +275,24 @@ module Pod
       end
     end
 
+    # @!group Updating the source
+    #-------------------------------------------------------------------------#
+
+    # Updates the local clone of the source repo.
+    # @return  [Array] output and changed_spec_paths
+    #         First value of the array is `git pull` output. Second value is an array of changed spec paths.
+    #
+    def update
+      output = nil
+      changed_spec_paths = nil
+      Dir.chdir(repo) do
+        prev_commit_hash = (`git rev-parse HEAD`).strip
+        output = `git pull --ff-only`
+        changed_spec_paths = (`git diff --name-only #{prev_commit_hash}..HEAD`).strip.split("\n")
+      end
+      [output, changed_spec_paths]
+    end
+
     public
 
     # @!group Representations
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -285,18 +285,11 @@ module Pod
     # @return  [Array<String>] changed_spec_paths
     #          Returns the list of changed spec paths.
     #
-    def update(show_output = false)
+    def update(show_output)
       changed_spec_paths = []
       Dir.chdir(repo) do
         prev_commit_hash = git_commit_hash
-        output = update_git_repo
-        CoreUI.puts output if show_output
-        unless $?.success?
-          CoreUI.warn 'CocoaPods was not able to update the ' \
-                  "`#{name}` repo. If this is an unexpected issue " \
-                  'and persists you can inspect it running ' \
-                  '`pod repo update --verbose`'
-        end
+        update_git_repo(show_output)
         changed_spec_paths = diff_until_commit_hash(prev_commit_hash)
       end
       changed_spec_paths
@@ -373,8 +366,9 @@ module Pod
       (`git rev-parse HEAD` || '').strip
     end
 
-    def update_git_repo
-      `git pull --ff-only 2>&1`
+    def update_git_repo(show_output = false)
+      output = `git pull --ff-only 2>&1`
+      CoreUI.puts output if show_output
     end
 
     def diff_until_commit_hash(commit_hash)
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -287,7 +287,8 @@ module Pod
       changed_spec_paths = nil
       Dir.chdir(repo) do
         prev_commit_hash = (`git rev-parse HEAD`).strip
-        output = `git pull --ff-only`
+        output = `git pull --ff-only 2>&1`
+        raise Informative, output unless $?.success?
         changed_spec_paths = (`git diff --name-only #{prev_commit_hash}..HEAD`).strip.split("\n")
       end
       [output, changed_spec_paths]
--- a/lib/cocoapods-core/source/aggregate.rb
+++ b/lib/cocoapods-core/source/aggregate.rb
@@ -53,7 +53,7 @@ module Pod
       #         The name of the Pod.
       #
       # @return [Set] The most representative set for the Pod with the given
-      #         name.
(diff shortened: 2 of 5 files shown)


lib/cocoapods-core/source/aggregate.rb
```diff
@@ -198,26 +164,34 @@ def update_search_index(search_data)
...
+            string << ' ' << k if k
+            string << ' ' << v if v
+          end
+        end
+        string.gsub(/\s+/m, ' ').strip.split(' ').uniq
       rescue
         CoreUI.warn "Skipping `#{set.name}` because the podspec contains " \
           'errors.'
-        result
+        return []
```
segiddins: `return` keyword is superfluous

lib/cocoapods-core/source/aggregate.rb
```diff
@@ -140,46 +140,12 @@ def search_by_name(query, full_text_search = false)
...
-          has_data = set_data && set_data['version']
-          if has_data
-            stored_version = Version.new(set_data['version'])
-            if stored_version < set.highest_version
-              search_data[set.name] = search_data_from_set(set)
-            end
-          else
-            search_data[set.name] = search_data_from_set(set)
+          word_list_from_set(set).each do |w|
+            result[w] = [] unless result[w]
```
segiddins: `(result[w] ||= []).push(set.name)`

spec/source/aggregate_spec.rb
```diff
@@ -104,50 +104,59 @@ module Pod
...
+        index['Full'].should == ['BananaLib']
+        index['of'].should == ['BananaLib']
+        index['chunky'].should == ['BananaLib']
+        index['bananas.'].should == ['BananaLib']
+        index['Corp'].should == ['BananaLib']
+        index['Monkey'].should == ['BananaLib']
+        index['Boy'].should == ['BananaLib']
+      end
+
+      # TODO: Enable and change test methods for updating index after incremental update completed.
```
segiddins: this will need to be addressed

lib/cocoapods-core/source/aggregate.rb
manuyavuz: I like this!

spec/source/aggregate_spec.rb
manuyavuz: Same for me. It was just to see if it works and to open it to discussion.

segiddins: cool, just adding the note so it we don't accidentally merge

spec/source/aggregate_spec.rb
```diff
@@ -98,56 +98,65 @@ module Pod
...
-      it 'deletes from the index the data of the sets which are not present in the aggregate' do
-        old_index = { 'Deleted-Pod' => { 'version' => '1.0', 'summary' => 'custom' } }
-        index = @aggregate.update_search_index(old_index)
-        index['Deleted-Pod'].should.be.nil
-      end
+        index = @aggregate.generate_search_index_for_source(@test_source)
+        index['BananaLib'].should == ['BananaLib']
+        index['JSONKit'].should == ['JSONKit']
+        index['JSONSpec'].should == ['JSONSpec']
+        index['Faulty_spec'].should.nil?
```
segiddins: `should.be.nil`

manuyavuz:
Got an offense from RuboCop about using return while returning two parameters.

I think I have to use return in this case, so we can ignore it.

Or maybe I can return an array value including two parameters. Which makes more sense?

lib/cocoapods-core/source.rb
```diff
@@ -259,6 +275,24 @@ def fuzzy_search(query)
...
+    #
+    def update
+      output = nil
+      changed_spec_paths = nil
+      Dir.chdir(repo) do
+        prev_commit_hash = (`git rev-parse HEAD`).strip
+        output = `git pull --ff-only`
+        changed_spec_paths = (`git diff --name-only #{prev_commit_hash}..HEAD`).strip.split("\n")
+      end
+      return output, changed_spec_paths
```
segiddins: `[output, changed_spec_paths]`, no return

segiddins:
@manuyavuz are these PRs ready for final review?

manuyavuz:
It lacks specs but I think we should be in the same line about the implementation before starting specs. So I think it's ready for review regarng implementation. What do you think?

lib/cocoapods-core/source.rb
```diff
@@ -259,6 +275,33 @@ def fuzzy_search(query)
...
+
+    # Updates the local clone of the source repo.
+    #
+    # @param  [Bool] show_output
+    #
+    # @return  [Array<String>] changed_spec_paths
+    #          Returns the list of changed spec paths.
+    #
+    def update(show_output = false)
+      changed_spec_paths = nil
```
segiddins: this should be `[]`

lib/cocoapods-core/source/aggregate.rb
```diff
@@ -198,26 +187,34 @@ def update_search_index(search_data)
...
+        if spec.description
+          string << ' ' << spec.description
+        end
+        if spec.authors
+          spec.authors.each_pair do |k, v|
+            string << ' ' << k if k
+            string << ' ' << v if v
+          end
+        end
+        string.gsub(/\s+/m, ' ').strip.split(' ').uniq
```
segiddins: `string.split(/\s+/m).uniq`?

manuyavuz: I think we still need strip here. Consider  the following:

```
irb(main):029:0* x
=> "    asdfweqr    sadfew   sdf   "
irb(main):030:0> x.split(/\s+/m).uniq
=> ["", "asdfweqr", "sadfew", "sdf"]  //These are spaces
irb(main):031:0> x.strip.split(/\s+/m).uniq
=> ["asdfweqr", "sadfew", "sdf"]
```

manuyavuz: `string.split` is the best way. I'll stick to it.

manuyavuz:
If we're OK with the current state, I'll move on adding specs. Just let me know.

segiddins:
I am, but maybe @floere has some thoughts

lib/cocoapods-core/source/aggregate.rb
```diff
@@ -198,26 +187,34 @@ def update_search_index(search_data)
...
+      def word_list_from_set(set)
         spec = set.specification
-        result['version'] = spec.version.to_s
-        result['summary'] = spec.summary
-        result['description'] = spec.description
-        result['authors'] = spec.authors.keys.sort * ', '
-        result
+        string = set.name.dup
+        if spec.summary
+          string << ' ' << spec.summary
```
floere: In Ruby, it's generally better to avoid `string << ' ' << spec.summary` because with each `<<`, a String is created. Use arrays to `<<` the strings, then `join` once at the end.

Or better yet, in this case, only split where necessary (e.g. in summary) and append the result to the array using `+=`, then `uniq` at the end.

floere:
Only one comment, repeated here, refering to the latest commit where this code is located, `string << ' ' << spec.summary`:
In Ruby, it's generally better to avoid `string << ' ' << spec.summary` because with each `<<`, a String is created. Use arrays to `<<` the strings, then `join` once at the end.
Or better yet, in this case, only split where necessary (e.g. in summary) and append the result to the array using `+=`, then `uniq` at the end.

floere:
Happy to wait for the specs now 😊 Thanks for all your work!!!

manuyavuz:
Great hint @floere 👏🏼

I'm starting to specs 💥

manuyavuz:
Done with specs. You can check out and review now. 

I'll rewrite history a little to have a convenient history. Other than that, I think this is ready to go to master.

segiddins:
Looks amazing to me :+1:
But I also want to let @floere have one last look

lib/cocoapods-core/source.rb
```diff
@@ -326,6 +369,18 @@ def specs_dir
       end
     end
 
+    def git_commit_hash
+      (`git rev-parse HEAD` || '').strip
+    end
+
+    def update_git_repo
```
segiddins: we should override this in CocoaPods/CocoaPods to use executable, so the output is still streamed?

manuyavuz: I don't think this is needed because it's output is handled and streamed in `update` method above. 
Look at the lines between 292-300. It uses CoreUI which is handled by CocoaPods by default in `user_interface.rb` as far as I understand.

segiddins: No, the output here will no longer be streamed -- it will only be printed after the command is finished.

manuyavuz: You're right, this prints after command is finished. So what should we do? 

I think we're gonna remove `CoreUI.puts output if show_output` part in update method here in order not to have duplicate output. However, I didn't fully get how to override a method of Source class in cocoapods code (probably in sources_manager.rb but how)?

I'm not a native ruby person so probably I miss sth :)

manuyavuz:
I think I've found the way to override `update_git_repo` method in CocoaPods/CocoaPods in order to enable output streaming. Sample code:

```ruby
  class Source
    extend Executable
    executable :git

    alias_method :_original_update_git_repo, :update_git_repo

    def update_git_repo
      git %w(pull --ff-only)
    end
  end
```

I will add to CocoaPods/CocoaPods#4249 in next commit.

segiddins:
@manuyavuz no need to alias

manuyavuz:
It was really simple then :+1:
We also need to move UI.warn logic into here because Executable does not set $? so CoreUI.warn in here does not called when pull fails.

Regarding this, we may completely remove UI printing here, and handle it in sources_manager again, this time in above overridden method. What do you think?

segiddins:
Sure

lib/cocoapods-core/source/aggregate.rb
```diff
@@ -187,8 +165,19 @@ def update_search_index(search_data)
...
       #-----------------------------------------------------------------------#
 
-      # Returns the search related information from the most representative
-      # specification of the set following keys:
+      # Generates search data for given array of sets.
+      def generate_search_index_for_sets(sets)
+        result = {}
+        sets.each do |set|
+          word_list_from_set(set).each do |w|
+            (result[w] ||= []).push(set.name)
```
floere: Depending on the average number of words in a wordlist, using the set name/identifier as a symbol will use a bit less space. If it's only used to generate the file index, and not for storing it in memory (and search operations) it's not that crucial.

lib/cocoapods-core/source/aggregate.rb
```diff
@@ -198,26 +187,34 @@ def update_search_index(search_data)
...
+      def word_list_from_set(set)
         spec = set.specification
-        result['version'] = spec.version.to_s
-        result['summary'] = spec.summary
-        result['description'] = spec.description
-        result['authors'] = spec.authors.keys.sort * ', '
-        result
+        word_list = [set.name.dup]
+        if spec.summary
+          word_list += spec.summary.split
```
floere: Note that the default split only splits on whitespaces. We may need to be cleverer and for example also split on other characters. One idea is to `scan` instead of split. Try this example: `ruby -e 'p "Hello. This-is/an!Example".scan(/\w+/)'` (I am sorry for having suggested `split` earlier, only to now suggest `scan` – `split` also works, of course)

I believe we need a spec for this method, since the index data is so crucial for later searching.

floere:
Looks good – only had comments on the index creation. Thanks again for your work, @manuyavuz! :rocket:

lib/cocoapods-core/source/aggregate.rb
manuyavuz: @floere What about things like emails, URLs, or word 'Objective-C' for example?

Here is a sample output:

```
irb(main):003:0> "Objective-C code sample!".scan(/\w+/)
=> ["Objective", "C", "code", "sample"]
```

I'm not sure relying on \w would give ideal results because generally computer science vocabulary contains non-alphabetical characters in terms. Do you have any suggestions on this?

manuyavuz: Btw tried scan and see that it has a slightly better performance than using split (11 secs vs 13 secs for indexing from scratch) and I liked the approach. But I'm afraid we could lose some common terms with it.

floere: The `\w` was just an example – we can be as clever as we want to be when tokenizing, handling special cases like `Objective-C`. I'd like to leave it up to you as the main CLI search engine writer how clever exactly we want to be. A more conservative possibility would be to use `.scan(%r{[\w-]+})` with a liberal sprinkling of non-alphabetic characters added in addition to the `-`.

manuyavuz: I will definitely want to challenge cleverness of the current system with better tokenization methods, but I feel like it should be another enhancement issue, and we should go with current implementation.

I guess I detected a more critical issue. Current implementation won't work for multi word queries like this: `pod search --full network library`.

I should split query string to array and perform search for each query, and then get the intersect of the pod arrays. It will take some time

floere: I agree with you on moving tokenization to a later issue to get things moving.




I just assumed you were doing an intersection - which is why I suggested using symbols in the inverted index array values (for memory savings and more efficiency on the intersection).

manuyavuz:
> I believe we need a spec for this method, since the index data is so crucial for later searching.

Actually the new specs I've added implicitly verify this method's correctness. What type of checks you think needed more? Like specs having no description, or authors part? Or like checking if strings are splitted correctly? 

Could you share if you have edge cases in your mind which you feel does not being tested now? I can add them accordingly.

floere:
@manuyavuz Just checking how strings are split. Very simple, maybe something as simple as `a.b-c/d` -> `%w(a b c d)`. This will help us later find bugs (since people like their pods be found and will complain if they can't), and guard against unintended consequences to tokenization changes. Something which the implicit tests can't do very well.

manuyavuz:
@floere I've added a spec for `word_list_from_set` method. Please tell if anything feels wrong or redundant.

floere:
@manuyavuz All good! :+1: :rocket:

manuyavuz:
Assuming this is ready to ⛵️, should I wait this to be merged in order to finish the last commit in CocoaPods/CocoaPods#4249? You remember I was directed Core repo to my branch for specs to be passed. If we merge this, I can revert it and submit last commit, so that we see green lights from boss travis.

manuyavuz:
Added changelog.

segiddins:
:rocket: :ship:

PR #266, opened 2015-09-23
Change required activesupport version to >=4.0.2

kattrali:
When a user only has activesupport 3.2.15 installed, invoking `plugin` in a Podfile calls [`Hash.deep_stringify_keys`](http://apidock.com/rails/v4.0.2/Hash/deep_stringify_keys) causing a [crash](https://github.com/CocoaPods/CocoaPods/issues/3076).  `deep_stringify_keys` was first added to activesupport in version 4.0.2, but the base version required by core is 3.2.15. As noted in the comments on Cocoapods/Cocoapods#2307, the base activesupport version was held back to continue support for versions of Ruby before 2.0.0.

As of 76845c8, Cocoapods no longer supports Ruby versions less than 2.0, invalidating the reasoning for holding back the base version number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CocoaPods Core Changelog
 
+## Master
+
+##### Bug Fixes
+
+* Fixes crash when using plugins where activesupport 4 was not installed.  
+  [Delisa Mason](https://github.com/kattrali)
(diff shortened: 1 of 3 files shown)


kylef:
Thanks @kattrali :balloon:.

PR #274, opened 2015-10-09
[Podfile]add support to specify a pod's source when there are multi sources

Whirlwind:
I have three spec repos in my company, and they are same code, but different type. A source code repo, a framework repo with debug mode and a framework repo with release mode!!!

```
source "git@mycompany.com/framework-release.git"
source "git@mycompany.com/framework-debug.git"
source "git@mycompany.com/source.git"
pod "AFNetwork"
pod "SDWebImage"
```

So, I will get the AFNetwork in the `framework-release` and the SDWebImage in the `framework-release`.

But, now, I need that SDWebImage `source code`, I must write `pod "SDWebImage", :git=>"xxxx"`.

I think it is not good idea.

So, I hope this:
```
pod "SDWebImage", :source=>"mycompany-source"
```

The `:source` said that I will use the SDWebImage in the `git@mycompany.com/source.git`

What about this idea?

### DIFF
--- a/lib/cocoapods-core/dependency.rb
+++ b/lib/cocoapods-core/dependency.rb
@@ -19,6 +19,11 @@ module Pod
     #
     attr_accessor :external_source
 
+    # @return [String] a string describing the source where the pod should
+    #         be fetched.
+    #
+    attr_accessor :source
+
     # @return [Bool] whether the dependency should use the podspec with the
     #         highest know version but force the downloader to checkout the
     #         `head` of the source repository.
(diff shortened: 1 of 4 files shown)


segiddins:
This will need a bunch of tests, both here and in cocoapods proper. We also need to consider how transitive dependencies of these dependencies ought to be resolved?

segiddins:
@Whirlwind unless you're willing to add test coverage for this feature, both here and in CocoaPods/CocoaPods, I'm afraid we won't be able to accept this contribution :(

Whirlwind:
Please wait ...

Whirlwind:
@segiddins I had add a few unit tests to my branch, and everything is ok.

segiddins:
We'll also need tests added to CocoaPods/CocoaPods, where dependency resolution actually takes place

segiddins:
Closing because @efirestone took care of everything :)

PR #280, opened 2015-12-02
Don't Wrap Long Lines

psycotica0:
I have internal packages with unfortunately long names.
This, though, causes CocoaPods to generate lock files it can't read, since it
wraps in such a way as to not be valid YAML anymore.

### DIFF
--- a/lib/cocoapods-core/yaml_helper.rb
+++ b/lib/cocoapods-core/yaml_helper.rb
@@ -96,7 +96,7 @@ module Pod
         case value
         when Array      then process_array(value)
         when Hash       then process_hash(value, hash_keys_hint)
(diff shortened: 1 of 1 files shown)


segiddins:
Could you please add a test and a CHANGELOG entry for this? Thanks!

segiddins:
Merged manually via https://github.com/CocoaPods/Core/commit/0556730bfe49801ec2201e1af4c790e111135b3e, thanks!

psycotica0:
You beat me to it! My testing showed that the line breaks only caused issues when they were hash keys, the list values worked fine. They both wrapped, but YAML / YAML Helper could reconstruct given the wrapping on list items. Anyway, thanks!

PR #286, opened 2015-12-29
Remove head dependencies

segiddins:


orta:
:+1:

mrackwitz:
:+1: LGTM

lib/cocoapods-core/dependency.rb
```diff
@@ -109,12 +104,8 @@ def initialize(name = nil, *requirements)
...
 
       elsif requirements.last == :head
-        @head = true
-        requirements.pop
-        unless requirements.empty?
-          raise Informative, 'A `:head` dependency may not specify version ' \
-            "requirements (#{name})."
-        end
+        raise Informative, '`:head` dependencies have been removed. Please use ' \
+          "normal external source dependencies instead of `:head` for `#{name}`."
```
manuyavuz: Should we improve this message a little bit more? I couldn't clearly understand what to use instead when I read the message here. Maybe we should include an example like  `:git => 'REPO_PATH'` inside the message?

manuyavuz:
:+1: other than the concern on informative message

lib/cocoapods-core/dependency.rb
orta: That's a great idea @manuyavuz - is it easy to access the `git` repo of the dependency at this point? We could provide a copy & pastable version, if not, I think ^ is :+1:

segiddins: no, this is just a `Dependency` object, it has no idea what the latest `Specification` will be

lib/cocoapods-core/version.rb
```diff
@@ -47,12 +47,9 @@ class Version < Pod::Vendor::Gem::Version
     #         A string representing a version, or another version.
     #
     def initialize(version)
-      if version.is_a?(Version) && version.head?
-        version = version.version
-        @head = true
-      elsif version.is_a?(String) && version =~ /HEAD based on (.*)/
+      if version.is_a?(String) && version =~ /HEAD based on (.*)/
```
segiddins: add `@todo` to remove this

PR #291, opened 2016-01-04
[Dependency] Allow creating from a string that has HEAD info

segiddins:
Closes CocoaPods/CocoaPods#4710.

\c @orta


orta:
Cool, makes sense

PR #292, opened 2016-01-07
Add ability to disable inhibiting warnings per pod

manuyavuz:
This enables setting `:inhibit_warnings => false` option to pods in podfile.
Warnings from pods having above value will not surpass even if `inhibit_all_warnings!` is set.

- [x] Specs
- [x] Changelog

### DIFF
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -286,7 +286,7 @@ module Pod
       def inhibits_warnings_for_pod?(pod_name)
         if Array(inhibit_warnings_hash['not_for_pods']).include?(pod_name)
           false
-        elsif raw_inhibit_warnings_hash['all']
+        elsif inhibit_warnings_hash['all']
           true
         elsif !root? && parent.inhibits_warnings_for_pod?(pod_name)
           true
@@ -650,7 +650,16 @@ module Pod
         if exclusive?
           inhibit_hash
         else
-          parent.send(:inhibit_warnings_hash).merge(inhibit_hash) { |l, r| (l + r).uniq }
+          parent_hash = parent.send(:inhibit_warnings_hash)
+          if parent_hash['not_for_pods']
+            parent_hash['not_for_pods'] -= inhibit_hash['for_pods']
+          end
+          if parent_hash['for_pods']
+            parent_hash['for_pods'] -= inhibit_hash['not_for_pods']
+          end
+          parent_hash.merge(inhibit_hash) { |_, l, r|
+            (l + r).uniq
+          }
         end
       end
 
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -318,12 +318,18 @@ module Pod
       # @return [void]
       #
       def set_inhibit_warnings_for_pod(pod_name, should_inhibit)
-        hash_key = 'for_pods' if should_inhibit == true
-        hash_key = 'not_for_pods' if should_inhibit == false
-        if hash_key
-          raw_inhibit_warnings_hash[hash_key] ||= []
-          raw_inhibit_warnings_hash[hash_key] << pod_name
-        end
+        hash_key = case should_inhibit
+                   when true
+                     'for_pods'
+                   when false
+                     'not_for_pods'
+                   when nil
+                     return
+                   else
+                     raise ArgumentError, "Got `#{should_inhibit.inspect}`, should be a boolean"
+                   end
+        raw_inhibit_warnings_hash[hash_key] ||= []
+        raw_inhibit_warnings_hash[hash_key] << pod_name
       end
 
       #--------------------------------------#
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -284,7 +284,9 @@ module Pod
       #         return true for any asked pod.
       #
       def inhibits_warnings_for_pod?(pod_name)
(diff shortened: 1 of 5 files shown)


lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -317,6 +319,18 @@ def inhibit_warnings_for_pod(pod_name)
...
       end
 
+      # Does not inhibit warnings for a specific pod during compilation even if inhibit_all_warnings! was set.
+      #
+      # @param  [String] pod_name
+      #         Name of the pod for which the warnings should not be suppressed.
+      #
+      # @return [void]
+      #
+      def do_not_inhibit_warnings_for_pod(pod_name)
```
segiddins: This should be merged with `inhibit_warnings_for_pod`, adding a bool parameter.

manuyavuz: definitely :) sorry for my bad

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -307,14 +309,21 @@ def inhibit_all_warnings=(flag)
...
+      # @param  [Bool] should_inhibit
+      #         Whether the warnings should be inhibited or not for given pod.
       #
       # @return [void]
       #
-      def inhibit_warnings_for_pod(pod_name)
-        raw_inhibit_warnings_hash['for_pods'] ||= []
-        raw_inhibit_warnings_hash['for_pods'] << pod_name
+      def set_inhibit_warnings_for_pod(pod_name, should_inhibit)
+        hash_key = 'for_pods' if should_inhibit == true
```
segiddins: `hash_key = should_inhibit ? 'for_pods' : 'not_for_pods'`

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -633,14 +642,24 @@ def raw_inhibit_warnings_hash
...
       #
       def inhibit_warnings_hash
         inhibit_hash = raw_inhibit_warnings_hash
         if exclusive?
           inhibit_hash
         else
-          parent.send(:inhibit_warnings_hash).merge(inhibit_hash) { |l, r| (l + r).uniq }
+          parent_hash = parent.send(:inhibit_warnings_hash)
+          if parent_hash['not_for_pods']
+            parent_hash['not_for_pods'] -= inhibit_hash['for_pods']
```
segiddins: this will mutate the parent, which is not what we want

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -633,14 +642,24 @@ def raw_inhibit_warnings_hash
...
+          end
+          if parent_hash['for_pods']
+            parent_hash['for_pods'] -= inhibit_hash['not_for_pods']
+          end
+          parent_hash.merge(inhibit_hash) { |_, l, r|
```
segiddins: block should stay on a single line

spec/podfile/dsl_spec.rb
```diff
@@ -101,6 +123,37 @@ module Pod
...
+            target 'App' do
+              pod 'PonyDebugger', :inhibit_warnings => true
+              target 'Inherited' do
+                pod 'PonyDebugger', :inhibit_warnings => false
+              end
+            end
+          end
+
+          target = podfile.target_definitions['Inherited']
+          target.inhibits_warnings_for_pod?('PonyDebugger').should.be.false
```
segiddins: `target.should.not.inhibits_warnings_for_pod('PonyDebugger')`

segiddins: please also test the app target here

segiddins:
Needs specs for overriding when the parent (or self) has `inhibit_all_warnings!` set.

lib/cocoapods-core/podfile/target_definition.rb
manuyavuz: If I do this, pod will inhibit for the cases user sets the value any object that is not boolean false or nil, such as `:inhibit_warnings => 'false'`. So I think we should explicitly check like this and ignore otherwise.

lib/cocoapods-core/podfile/target_definition.rb
manuyavuz: Hmm yeah, I think `parent_hash.dup` would be enough, do you see any other possible issues on that?

lib/cocoapods-core/podfile/target_definition.rb
segiddins: in that case, do
```ruby
hash_key = case should_inhibit
when true
 'for_pods'
when false
  'not_for_pods'
else
  raise ArgumentError, "got `#{should_inhibit.inspect}`, should be a boolean"
end
```

lib/cocoapods-core/podfile/target_definition.rb
segiddins: that should be fine, just make sure to test it

lib/cocoapods-core/podfile/target_definition.rb
manuyavuz: Cool, I still couldn't get used to some of Ruby's cool features :)

manuyavuz:
Wouldn't that case be tested with the spec here: https://github.com/CocoaPods/Core/pull/292/files#diff-0ac902131aac45fbb1a0c03822b97d01R80

Do you mean the case of overriding during inheritance while also having `inhibit_all_warnings!` on top?

segiddins:
@manuyavuz yes

segiddins:
:+1:

PR #299, opened 2016-01-26
[DSL] Rename xcodeproj to project

### LINKED ISSUE #298: DSL directive `project` instead of `xcodeproj`
* The directive to specify a project is called `xcodeproj`.
* The directive to specify a workspace is called `workspace`.
* The construct to declare a target is called `target`.

But all those are based on artifacts created with or for Xcode. I think that's one of the most counter-intuitive things in the current DSL. I'd love when we could fix that, before we "finalize" the DSL for 1.0.

mrackwitz:
Fixes #298.

* [x] Test for the deprecated method
* [x] Changelog entry
* [x] Needs a big sister PR adopting this change on the main repo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # CocoaPods Core Changelog
 
+## Master
+
+##### Breaking
+
+* The `xcodeproj` Podfile DSL method has been renamed to `project`.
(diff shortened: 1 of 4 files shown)


segiddins:
👍🏻

PR #301, opened 2016-01-29
[TargetDefinition] Safe access for bool values

mrackwitz:
`false` values were set to nil, when they were read, so that `use_frameworks?` would be evaluated to `true` on the next access.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@
   [Marius Rackwitz](https://github.com/mrackwitz)
   [Core#298](https://github.com/CocoaPods/Core/issues/298)
 
+##### Bug Fixes
+
+* Fix accessing `use_frameworks?` without accidentally clearing the value when
+  it was explicitly set to `false`, so that it would be evaluated as `true` on
+  the next access.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
   [Marius Rackwitz](https://github.com/mrackwitz)
   [Core#298](https://github.com/CocoaPods/Core/issues/298)
 
+##### Bug Fixes
+
+* Fix accessing `use_frameworks?` without accidentally clearing the value when
+  it was explicitly set to `false`, so that it would be evaluated as `true` on
+  the next access.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [Core#301](https://github.com/CocoaPods/Core/issues/301)
```
segiddins: 301 isn't an issue

spec/podfile/target_definition_spec.rb
```diff
@@ -294,6 +294,7 @@ module Pod
         @parent.use_frameworks!
         @child.use_frameworks!(false)
         @child.should.not.uses_frameworks?
+        @child.should.not.uses_frameworks?
```
segiddins: Please add a comment saying why this line is necessary, since it looks superfluous.

mrackwitz: :+1:

PR #303, opened 2016-01-30
[Dependency] Add ability to ignore requirements for `from_string`

manuyavuz:
Related to CocoaPods/CocoaPods#4846.

This is needed to unlock local dependencies during locking dependency graph generation.

- [ ] CHANGELOG
- [ ] Spec

### DIFF
--- a/lib/cocoapods-core/dependency.rb
+++ b/lib/cocoapods-core/dependency.rb
@@ -344,11 +344,13 @@ module Pod
     #
     # @return   [Dependency] the dependency described by the string.
     #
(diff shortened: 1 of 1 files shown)


segiddins:
Please no. I'd much rather, where this is needed, parse the string as normal and create a new dependency just using the name.

manuyavuz:
Are you sure? I don't want to add the same regex parsing logic into another place because we would need to change both parts if anytime lockfile format changes.

manuyavuz:
I'm closing this because it's not needed anymore.

PR #305, opened 2016-02-03
[Specification DSL] default_subspecs is only allowed at root level

mrackwitz:
… and some house-keeping.

* [x] Changelog


orta:
Good move

segiddins:
:+1: with changelog

PR #307, opened 2016-02-16
[PlatformProxy] Raise an error when calling platform=

segiddins:


segiddins:
Merging since no one had any feedback

PR #319, opened 2016-04-11
Clarify pod nesting + fix grammar

charliewilliams:


orta:
As an English native, you think I've be better at this it's business. Anyway, I think this is a good improvement - thanks! Gets my :+1:

orta:
The irony in that ^ comment having a typo is awesome

PR #325, opened 2016-05-11
Improve documentation for Podfile's 'install!' attribute

AliSoftware:
It was not clear to me when reading that what the first parameter was meant to be, if it was always meant to be `'cocoapods'` or if other values might be acceptable, etc, hence this more specific documentation.


segiddins:
👍🏻

orta:
aye - good job

PR #327, opened 2016-05-17
Improved warning message for DSL attribute renaming

AliSoftware:
A lot of my coworkers and friends saw the "xcodeproj was renamed to project" warning but thought it was a warning inside CocoaPods' own code, like some warning aimed at us and that we'd have to fix in the future.

They didn't understand that it was meant to suggest that **they** should update their `Podfile` to fix it. Hence that new message to make things clearer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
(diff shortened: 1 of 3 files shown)


orta:
👍 yep

PR #331, opened 2016-05-23
[TargetDefinition] Return nil when storing a pod

segiddins:
Closes https://github.com/CocoaPods/CocoaPods/issues/5379


orta:
Nice, yeah, this seems like a good idea 👍

PR #335, opened 2016-06-14
Fix slow dependency resolution

yanzhiwei147:
For master issue [#5180](https://github.com/CocoaPods/CocoaPods/issues/5180)

### DIFF
--- a/lib/cocoapods-core/specification.rb
+++ b/lib/cocoapods-core/specification.rb
@@ -36,6 +36,7 @@ module Pod
       @consumers = {}
       @parent = parent
       attributes_hash['name'] = name
+      @hash_value = 0
 
       yield self if block_given?
     end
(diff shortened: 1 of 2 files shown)


lib/cocoapods-core/specification.rb
```diff
@@ -80,7 +81,10 @@ def ==(other)
     # @return [Fixnum] The hash value.
     #
     def hash
-      (name.hash * 53) ^ version.hash
+      if @hash_value == 0
```
segiddins: why not `@hash ||=`

segiddins: you'll also need to reset the hash when changing the name or the version

segiddins:
What sort of performance improvements does this give?

yanzhiwei147:
@segiddins I found it calculate hash value more than 10 billion when run the `pod update `(It access frequently).

segiddins:
Right, but could you please share a benchmark or how much time these changes save?

yanzhiwei147:
I know the total save time only, the same as this [PR](https://github.com/CocoaPods/CocoaPods/pull/5510).

segiddins:
If it's a 4x speedup, great! Can you please add a changelog entry and address the outstanding comments? Thanks!

yanzhiwei147:
Yes, I can add the changelog.

PR #339, opened 2016-06-28
[docs] example on main Podfile section is invalid

maschall:
🌈

There is a discrepancy between the example for post_hook and the "complex" example on the main Podfile Syntax Reference


segiddins:
Thanks @maschall !

PR #350, opened 2016-08-09
Improved handling of Semver pre-release versions

benasher44:
Fixes CocoaPods/CocoaPods#5718

- Handle special case of missing first segment of pre-release segments (doesn't mean it wasn't handled before, just calling out this part in the diff)
- Updated according to [Semver's predence spec](http://semver.org/#spec-item-11) (item 11 if it doesn't link you straight there):

> Numeric identifiers always have lower precedence than non-numeric identifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Improved comparison between Semver pre-release versions.  
+  [Ben Asher](https://github.com/benasher44)
+  [#350](https://github.com/CocoaPods/Core/pull/350)
 
 ##### Bug Fixes
 
--- a/lib/cocoapods-core/version.rb
+++ b/lib/cocoapods-core/version.rb
@@ -194,22 +194,33 @@ module Pod
       return unless other.is_a?(Pod::Version)
       return 0 if @version == other.version
 
-      compare = proc do |segments, other_segments|
+      compare = proc do |segments, other_segments, is_pre_release|
         limit = [segments.size, other_segments.size].max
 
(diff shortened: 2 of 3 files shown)


segiddins:
amazing ❤️ 

:shipit: with a CHANGELOG

benasher44:
@segiddins what's the difference between `Pod::Version` and `Pod::Vendor::Version`? Should I add test cases for that one as well?

segiddins:
@benasher44 `Pod::Version` is a subclass of the vendored RubyGems implementation, so that's not necessary :)

benasher44:
👍

benasher44:
@segiddins wait no I'm still confused. `Pod::Version` inherits from `Pod::Vendor::Gem::Version`, and `Pod::Vendor::Gem::Version` also implmements `<=>`. `Pod::Vendor::Gem::Version`'s implementation of `<=>` looks how `Pod::Version`'s used to look, so it should be prone to the same bug no?

segiddins:
but `Pod::Vendor::Gem::Version` is never used directly, only via `Pod::Version`

benasher44:
Huh so it's not. Looks like we only ever refer to `Gem::Version`. Any reason not to collapse `Pod::Vendor::Gem::Version` into `Pod::Version`?

segiddins:
To make it easier to merge in changes from upstream

benasher44:
OH! This is just a clone of [Gem::Version](/rubygems/rubygems/blob/master/lib/rubygems/version.rb). Welp. This all makes sense now. Thanks!

segiddins:
🚢

PR #359, opened 2016-11-29
Update smart quotes warning

allewun:
I was confused by the original messaging. To me, it implied that the `Podfile` would be saved with the sanitization, which did not happen. Not sure this how others might read it, so open to suggestions here.


orta:
Maybe "detected and ignored"? other than that, seems like a nice idea - 👍

allewun:
Done, only took 2 months to add 2 words 😄

DanToml:
@allewun 🎉  thank you!

PR #372, opened 2017-04-21
Clarify the existance of public, private and project headers

### LINKED ISSUE #371: Podspec Guide should mention project header files
(Opening the issue here because, if I understand it right, the Guide is generated from a file in this repo)

Currently, the [Podspec Guide](https://guides.cocoapods.org/syntax/podspec.html) provides description of ```public_header_files``` and ```private_header_files```. This creates an impression that there are two kinds of header files, and each header file in the pod's sources will end up being either public or private.

At the same time, actually there is a third type - header files that do not appear in neither of the above. For instance, in the generated project file, these files [will be added to the "Project" section of the "Headers" build phase](https://github.com/CocoaPods/CocoaPods/blob/8df7c49a783ab1661e8e1b532f7725a1604c2c39/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb#L403), unlike public and private header files.

Suggested fix: add a short note about the "project header" files, although there are no really such field in the podspec.

happycoloredbanana:
Fixes #371.

The pod creator should be aware of [what will happen to the headers files in Xcode build phases](https://github.com/CocoaPods/CocoaPods/blob/8df7c49a783ab1661e8e1b532f7725a1604c2c39/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb#L403).

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1012,8 +1012,10 @@ module Pod
       #   ---
       #
       #   These are the headers that will be exposed to the user’s project and
-      #   from which documentation will be generated. If no public headers are
-      #   specified then **all** the headers in source_files are considered public.
+      #   from which documentation will be generated. When the library is built,
+      #   these headers will appear in the build directory. If no public headers
+      #   are specified then **all** the headers in source_files are considered
+      #   public.
       #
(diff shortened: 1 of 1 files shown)


orta:
👍 works for me, let's do it - thanks @happycoloredbanana

PR #313, opened 2016-03-09
Check the GitHub API before Fetching the Master Specs repo

DanielTomlinson:
- [x] Add MasterSource
- [x] Refactor Source::Aggregate to be initialised with Array<Source>
- [x] Update CocoaPods/CocoaPods SourcesManager for new Aggregate

As recommended https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193801376

Supersedes https://github.com/CocoaPods/CocoaPods/pull/5001

CocoaPods/CocoaPods counterpart: https://github.com/CocoaPods/CocoaPods/pull/5014

PR #374, opened 2017-05-14
Apply GitHub checking logic to all GitHub hosted sources

greysteil:
Applies the GitHub efficiency check added in https://github.com/CocoaPods/Core/pull/313 to any spec repo which is hosted on GitHub.

Wanted to check approach before I add specs (i.e., move them from the ones currently on `MasterSource`). Are you guys OK with the above in principle, and do you see any reason for the continuing existence of `MasterSource` once it's done?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ coverage/
 .coveralls.yml
 .idea/
 documentation/
+.bundle/
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,3 +33,8 @@ before_install:
         git remote add origin https://github.com/CocoaPods/Specs.git
       )
       git submodule update --init
+      (
+        cd spec/fixtures/spec-repos/test_repo
+        git init
+        git remote add origin https://bitbucket.com/test/test_repo.git
+      )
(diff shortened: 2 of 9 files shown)


orta:
I _think_ one of the original reasons here is that some enterprises installs of github may not have the API for checking, but in this case you are still only checking on github.com - so that's a pretty good safe bet still.

WRT the MasterSource - I'll leave that answer to someone who is more familiar with the codebase

greysteil:
Thanks @orta. Anyone got a view on the MasterSource? Seems sensible to get rid of it if it's not doing anything...

orta:
@segiddins you probably have the most long term perspective on this

segiddins:
MasterSource is there so in the future we could apply separate logic that doesn't depend on the file system

greysteil:
@segiddins - awesome, I've added a comment on the now empty class saying as such.

Specs are in and all passing. Had to make some tweaks to the GitHub regex, but I'm pretty sure they're improvements. What do you reckon?

dnkoutso:
LGTM. I'll merge and ask for forgiveness if this needs additional work or needs to be reverted.

greysteil:
❤️

PR #375, opened 2017-05-22
Allow ActiveSupport 5+

greysteil:
Having looked through the previous conversation on locking ActiveSupport in CocoaPods ([here](https://github.com/CocoaPods/CocoaPods/pull/5602)) I can't see any reason why it should be locked in Core, too. Specs were previously failing on `2.0.0-p481`, which had a buggy version of Bundler on Travis (presumably the same buggy version which was causing the issue on clean installs in the first place), but I think it should be fine to unlock the gem here and just fix the spec setup. Won't affect CocoaPods since that has a tighter specification in its gemspec.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ script: bundle exec rake spec
 git:
   submodules: false
 before_install:
+  - gem install bundler
   - |
       master_sha="$(ruby -e "puts '`git submodule status spec/fixtures/spec-repos/master`'.strip.sub(/^-/, '').split(/\s/, 2).first")"
       curl -ssL "https://github.com/CocoaPods/Specs/archive/$master_sha.tar.gz" | tar xz -C spec/fixtures/spec-repos
--- a/cocoapods-core.gemspec
+++ b/cocoapods-core.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.files         =  Dir["lib/**/*.rb"] + %w{ README.md LICENSE }
   s.require_paths =  %w{ lib }
 
-  s.add_runtime_dependency 'activesupport', '>= 4.0.2', '< 5'
+  s.add_runtime_dependency 'activesupport', '>= 4.0.2', '< 6'
   s.add_runtime_dependency 'nap', '~> 1.0'
   s.add_runtime_dependency 'fuzzy_match', "~> 2.0.4"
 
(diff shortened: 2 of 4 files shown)


greysteil:
@DanToml - does this look OK to you? I'm guessing from https://github.com/CocoaPods/Core/commit/7d909cd3980ce12fb1ecc482225e3c3fabbda9fe that you're the most expert on this bit of the codebase!

segiddins:
Please don't remove the lockfile! I believe that allowing AS 5 here should be fine, as long as we keep the stricter requirement in the CocoaPods gem spec (and we should probably pin to AS 4.x in the gemfile for testing purposes, as that's what CocoaPods users will be using)

greysteil:
Done. Out of curiosity, what's the reason for the Gemfile.lock?

segiddins:
It means we can reproduce test outcomes deterministically

cocoapods-core.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.files         =  Dir["lib/**/*.rb"] + %w{ README.md LICENSE }
   s.require_paths =  %w{ lib }
 
-  s.add_runtime_dependency 'activesupport', '>= 4.0.2', '< 5'
+  s.add_runtime_dependency 'activesupport', '>= 4.0.2'
```
segiddins: probably should keep an upper bound of 6 for the time being

.travis.yml
```diff
@@ -22,6 +22,7 @@ script: bundle exec rake spec
 git:
   submodules: false
 before_install:
+  - gem update bundler
```
segiddins: please use `gem install` instead of `gem update`

greysteil:
Updated!

Would love your thoughts on https://github.com/CocoaPods/CocoaPods/pull/5602 if you have a moment, too - I'd really like to get CocoaPods working with Rails 5.

greysteil:
@segiddins - any chance of getting this merged?

segiddins:
@DanToml thoughts?

DanToml:
This should be fine for Core with a rebase, esp while it is still locked in CocoaPods/CocoaPods.

@greysteil could you open a new issue on CocoaPods/CocoaPods rather than discussing in the old PR? there's a lot more thought to be done there because we have to support macOS system rubygems.

greysteil:
Sure thing. Rebased and everything here is 💚  so hopefully good to go?

DanToml:
@greysteil thank you! - I hope your project goes well 💜

PR #386, opened 2017-06-17
Static framework

paulb777:
Introduce static_framework podspec attribute to indicate that a source CocoaPod should be built into a static library framework. 

More details in the corresponding CocoaPods PR - coming shortly.

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -458,6 +458,25 @@ module Pod
 
       #------------------#
 
+      # @!method static_framework=(flag)
+      #
+      #   Indicates, that if use_frameworks! is specified, the
+      #   pod should include a static library framework.
+      #
+      #   @example
+      #
+      #     spec.static_framework = true
+      #
+      #   @param [Bool] flag
+      #          Indicates, that if use_frameworks! is specified, the
+      #          pod should include a static library framework.
+      #
+      root_attribute :static_framework,
+                     :types => [TrueClass, FalseClass],
+                     :default_value => false
+
+      #------------------#
+
       # @!method deprecated=(flag)
       #
       #   Whether the library has been deprecated.
--- a/lib/cocoapods-core/specification/root_attribute_accessors.rb
+++ b/lib/cocoapods-core/specification/root_attribute_accessors.rb
@@ -152,6 +152,13 @@ module Pod
           command.strip_heredoc.chomp if command
         end
 
+        # @return [Bool] Indicates, that if use_frameworks! is specified, the
+        #         framework should include a static library.
+        #
+        def static_framework
+          attributes_hash['static_framework']
+        end
+
         # @return [Bool] Whether the Pod has been deprecated.
         #
         def deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Introduce `static_framework` podspec attribute
+  [Paul Beusterien](https://github.com/paulb777)
+  [#386](https://github.com/CocoaPods/Core/pull/386)
 
 ##### Bug Fixes
 
--- a/spec/specification/dsl_spec.rb
+++ b/spec/specification/dsl_spec.rb
@@ -76,6 +76,11 @@ module Pod
         @spec.attributes_hash['prepare_command'].should == 'ruby build_files.rb'
       end
 
+      it 'allows to specify whether the Pod has a static_framework' do
+        @spec.static_framework = true
+        @spec.attributes_hash['static_framework'].should == true
+      end
+
       it 'allows to specify whether the Pod has been deprecated' do
         @spec.deprecated = true
         @spec.attributes_hash['deprecated'].should == true
--- a/spec/specification/root_attribute_accessors_spec.rb
+++ b/spec/specification/root_attribute_accessors_spec.rb
@@ -143,6 +143,11 @@ module Pod
       @spec.prepare_command.should == 'ruby prepare_script.rb'
     end
 
+    it 'returns whether the Pod should build a static framework' do
+      @spec.static_framework = true
+      @spec.static_framework.should == true
+    end
+
     it 'returns whether the Pod has been deprecated' do
       @spec.deprecated = true
       @spec.deprecated.should == true


paulb777:
The corresponding CocoaPods PR is https://github.com/CocoaPods/CocoaPods/pull/6811

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -463,6 +463,25 @@ module DSL
...
+      #   Indicates, that if !use_frameworks is specified, the
+      #   pod should include a static library framework.
+      #
+      #   @example
+      #
+      #     spec.static_framework = true
+      #
+      #   @param [Bool] flag
+      #          Indicates, that if !use_frameworks is specified, the
+      #          pod should include a static library framework.
```
dnkoutso: what does "include" mean in this case? Does it mean "if use_frameworks! is specified then this pod will be packaged as a static framework?

Also here and throughout `!use_frameworks` should be `use_frameworks!`

dnkoutso:
@paulb777 first of all thanks for the PRs and effort to improve the project. I'd like to see the opinion of others around this but one of my bigger aspirations for CocoaPods packaging is to push it down further to the consumers.

For example, I had imagined a `link_style` attribute in which every `pod` line in our `Podfile` would provide a hint (or sometimes requirement in the case of Swift) to be packaged in a specific way.

See how Facebooks Buck has its documentation (https://buckbuild.com/rule/apple_library.html) for something similar.

The idea is that different pods can be packaged and linked differently but that would be up to the consumer of them. This would essentially unlock mixing frameworks with static libraries in CocoaPods and could also yield automatic static library diamond dependencies (in which a static lib is linked into two or more frameworks which would lead to duplicate symbols).

I don't think this should be part of the podspec, sources are sources and how they are packaged is a different task in my mind at least.

Thoughts?

dnkoutso:
From a perspective to merge this I am only hesitant because those values will be serialized and we will have to live with them forever. So we need to carefully think about this. I think your PR looks good and I am supportive of such progress but would like us to consider a long term path for supporting these different configurations.

dnkoutso:
Perhaps we can introduce the `link_style` attribute from this PR and let `use_frameworks!` drive the value for each `pod` line. So having:

```ruby
use_frameworks!
target 'MyApp'
  pod 'MyPodA' # :link_style => :dynamic, implied by the use_frameworks! flag for now.
end
```

paulb777:
@dnkoutso Thanks for the quick response! In general, I agree with the philosophy of pushing the control to the library consumer and I started with that approach. However, after thinking deeper about it, I decided the podspec approach was better for several reasons:

- The podspec approach is much simpler, but sufficient to meet our needs for Firebase - and likely for many others currently restricted to delivering static library frameworks via vendored_frameworks.

- This does not change CocoaPods current functionality. Library vendors can already deliver static frameworks - they just have to build it outside of CocoaPods. 

- The pod supplier will know if a library needs to be static or not, while the nuances of static versus dynamic will be lost on the vast majority of consumers.

- This is a low risk approach. The implementation has no impact on libraries and dependency management unless the new option is used. When the option is used, the implementation strives to treat it exactly like vendored_framework static frameworks.

- An automatic approach could solve the case of wanting a static framework because of dependencies, but not the case of wanting a static framework for startup performance reasons.

- The only reason I can think of to build a low-level Objective C libraries dynamically would be to support Swift client libraries (until Swift stabilizes and supports static).

- In my experience with Bazel, consumer libraries and apps have no control over how the libraries they're linking are built. All of the specifications are on the libraries themselves.

- I haven't used Buck, but this PR looks to be an implementation of preferred_linkage=static. preferred_linkage=any could be a future enhancement. link_style sounds like it correlates with use_frameworks!.  Note that both of these are library options (Podspec) and not app options (Podfile).

dnkoutso:
@paulb777 I haven't forgotten about this, will need a bit of time to think about it, in the meantime I encourage others to weigh in.

paulb777:
Thanks @dnkoutso 

Let me know if I can do anything to help reach a resolution. Would a conversation/meeting be useful?

orta:
Hey, sorry for the long delay for another opinion from our side.

I agree with your comments in https://github.com/CocoaPods/Core/pull/386#issuecomment-309282234 - this does feel like the lowest impact way to support this. I have been out of the Cocoa build toolchain world long enough to not necessarily be able to have a really long-term perspective on it (as I do know it's [kinda unsupported](https://twitter.com/rsebbe/status/878533316022067200)). 

However: I think this is probably something that people will always want, and I think this is probably the right way to do it.

paulb777:
Thanks for the review @orta!  

Getting these two PR's in will be really helpful for us at Google in [open sourcing the build process](https://github.com/CocoaPods/CocoaPods/issues/6772) for our [Firebase SDK](https://github.com/firebase/firebase-ios-sdk).  

The feature will also be helpful to anyone who wants to use a CocoaPods development process for static library framework pods. They won't have to develop and maintain external scripts to run xcodebuild to construct fat static library frameworks. Instead they can just deliver source and leave the build to the app developer.

paulb777:
@dnkoutso - what's the right way to make progress on this PR?

dnkoutso:
@paulb777 sorry I was out on vacation for the past 2 weeks. I think we should land this. Can you rebase this branch?

DanToml:
I somewhat agree with @orta, this feels like the right way to make the change, but I think I'd be a -1 until it was officially supported by Xcode.

paulb777:
@dnkoutso - I've done the rebase

@DanToml - I understand the concern.  It is annoying that Xcode doesn't have first class support for static frameworks. 

However, many people use CocoaPods with manually constructed static frameworks and this PR removes that error-prone manual step for library creators and makes it possible for other library creators to depend on their libraries. 

In addition, this PR makes it a lot easier for library creators to share buildable source (as opposed to shipping their static libraries as vendored_frameworks). This is very much in the spirit of CocoaPods and Apple's directions of encouraging source libraries.

paulb777:
@dnkoutso @DanToml What's the next step?

dnkoutso:
@paulb777 not up to me 100%. I think others have a concern about merging this. 

I am fairly new in the contributors list so I'd like not to make the call :)

orta:
We had a chat and think this is 👍 

Sorry it took so long, but we really do try to have a very long-term perspective on these things. CocoaPods is a relatively slow moving project because of it's maturity and team size to user base ratio. 

So, making changes about something that has the potential to change just ~260k apps from one dep the next time they update required a bit of a nuanced conversation. Most PRs are not like this 💃 

If this passes review, it's good to merge.

paulb777:
Squashed

lib/cocoapods-core/specification/root_attribute_accessors.rb
```diff
@@ -146,6 +146,20 @@ def prepare_command
...
+        #         framework should include a static library.
+        #
+        def static_framework?
+          attributes_hash['static_framework']
+        end
+
+        # @return [Bool] Indicates, that if use_frameworks! is specified, the
+        #         framework should include a static library.
+        #
+        def static_framework
```
dnkoutso: is this method needed? It seems like a dup of the one above?

dnkoutso:
LGTM. @paulb777 we are getting ready to do a 1.3.0 release with test specs, I think this would move to 1.4.0 as a fast follow release with a few other things.

lib/cocoapods-core/specification/root_attribute_accessors.rb
paulb777: My ruby is too weak to understand the reasons, but I get unit test failures like` ["[ERROR] [attributes] Unable to validate 'static_framework' (undefined method 'static_framework' for #<Pod::Specification name=nil>\nDid you mean?  static_framework?\n               static_framework=)."] ` if I remove it.

dnkoutso: It comes from:

```ruby
        def value_for_attribute(attribute)
          if attribute.root_only?
            consumer.spec.send(attribute.name)
          else
            consumer.send(attribute.name) if consumer.respond_to?(attribute.name)
          end
        rescue => e
          results.add_error('attributes', "Unable to validate `#{attribute.name}` (#{e}).")
          nil
        end
```

It will automatically try to invoke a method with the name of the attribute to receive the value.

dnkoutso: I think you can remove the `?` method instead.

orta:
This feature is now officially supported by Xcode:

- https://twitter.com/SmileyKeith/status/889533496808615939
-https://twitter.com/daniel_dunbar/status/889533883338838016

dnkoutso:
Nice! We should definitely land this for 1.4.0.

lib/cocoapods-core/specification/root_attribute_accessors.rb
```diff
@@ -146,6 +146,13 @@ def prepare_command
           command.strip_heredoc.chomp if command
         end
 
+        # @return [Bool] Indicates, that if use_frameworks! is specified, the
+        #         framework should include a static library.
+        #
+        def static_framework
```
segiddins: since this returns a bool, maybe this should be `static_framework?`

paulb777:
Should there be both `static_framework` and `static_framework?`. 

See the discussion above about a test failures if `static_framework` is removed.

dnkoutso:
@paulb777 should we continue down that path or just instead update cocoapods to allow `.swift` files within static library targets based on xcode 9 b4 update?

paulb777:
@dnkoutso It still needs to be verified, but the static framework changes I made should wrap static libraries whether they're built from Objective-C or Swift.

keith:
I think we're talking about 2 separate issues here. This PR appears to be adding a podspec attribute `static_framework` a'la `vendored_framework`, where @dnkoutso (and I) are interested in the new static library/framework with Swift sources support that is officially supported as of Xcode 9 beta 4. In the latter case, CocoaPods would actually need to create static framework/library targets with Swift sources, something that we'll definitely need changes for.

keith:
It looks like we're now tracking the latter issue here: https://github.com/CocoaPods/CocoaPods/issues/6899

DanToml:
@paulb777 Could you rebase this PR then it's good to merge now 1.3.0 is out

paulb777:
@DanToml Rebased. Thanks!

ElizaSapir:
@paulb777  hey,

Having similar issue with `GVRKit`, I added `GVRKit` as a dependency on my podspec,
When trying to import the module get below issue:
`module 'GVRKit' not found`

Maybe you have some advise for me the only solution that works for me is:
simply set the framework search path to $(SRCROOT) and set it to recursive.

BR,
Eliza

paulb777:
Hey @ElizaSapir I don't see enough context to know what's going on.  I'd recommend creating a new issue and sharing a reproducible example.

PR #393, opened 2017-07-13
Check requires arc true false strings

VinayGuthal:
This fixes https://github.com/CocoaPods/CocoaPods/issues/6701.
Add a warning if the strings` 'true'` or `'false' ` are being used instead of boolean `true` or `false`  as the value for 'requires_arc' in the pod spec file.

### DIFF
--- a/lib/cocoapods-core/specification/linter.rb
+++ b/lib/cocoapods-core/specification/linter.rb
@@ -48,6 +48,7 @@ module Pod
         if spec
           validate_root_name
           check_required_attributes
+          check_requires_arc_attribute
           run_root_validation_hooks
           perform_all_specs_analysis
         else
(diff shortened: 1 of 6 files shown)


dnkoutso:
@VinayGuthal please add `CHANGELOG` entry, run `rubocop -a` and squash to a single commit.

LGTM.

lib/cocoapods-core/specification/linter.rb
```diff
@@ -99,6 +100,22 @@ def validate_root_name
         end
       end
 
+      # Generates a warning if the requires_arc attribute has true or false string values.
+      #
+      # @return [void]
+      #
+      def check_requires_arc_attribute
+        attributes = DSL.attributes.values
+        attributes.each do |attr|
```
dnkoutso: can you just use `attributes.find { |attr| attr.name == :requires_arc }` instead?

lib/cocoapods-core/specification/linter.rb
```diff
@@ -99,6 +100,20 @@ def validate_root_name
         end
       end
 
+      # Generates a warning if the requires_arc attribute has true or false string values.
+      #
+      # @return [void]
+      #
+      def check_requires_arc_attribute
+        attribute = DSL.attributes.values.find{|attr| attr.name == :requires_arc}
```
dnkoutso: is the formatting correct here? I think spaces are needed, `rubocop -a` I believe will take care of it for you.

dnkoutso:
👌 !

@DanToml or @orta quick check?

orta:
Yep, good idea

VinayGuthal:
Is there anything else that I need to take care of here? :)

orta:
No, I don't think so, let's get this in.

PR #401, opened 2017-08-29
Fix requirements cloning for `:testspecs`

justinseanmartin:
🌈

Found an issue where the requirements weren't being deep copied when using the `:testspecs` directive in combination with the `:configurations` directive. It would lead to an error message showing up with:

> [!] The subspecs of `MyPod` are linked to different build configurations for the `Pods-MyApp` target. CocoaPods does not currently support subspecs across different build configurations.

This addresses the issue by deep copying the requirements Hash contained within the Array instead of just copying the array containing those requirements.

### DIFF
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -921,7 +921,8 @@ module Pod
         end if subspecs
 
         test_specs.each do |ss|
-          store_pod("#{name}/#{ss}", *requirements.dup)
+          requirements_copy = requirements.map(&:dup)
+          store_pod("#{name}/#{ss}", *requirements_copy)
         end if test_specs
 
         requirements.pop if options.empty?
--- a/spec/podfile/target_definition_spec.rb
+++ b/spec/podfile/target_definition_spec.rb
@@ -409,6 +409,15 @@ module Pod
         @child.all_whitelisted_configurations.sort.should == %w(Debug Release)
       end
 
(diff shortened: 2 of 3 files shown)


dnkoutso:
We discovered this locally, verified it works as expected.

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -921,7 +921,8 @@ def parse_subspecs(name, requirements)
         end if subspecs
 
         test_specs.each do |ss|
-          store_pod("#{name}/#{ss}", *requirements.dup)
+          requirements_copy = requirements.map(&:dup)
+          store_pod("#{name}/#{ss}", *requirements_copy)
```
dnkoutso: this was basically returning a shallow clone of the requirements and when using test specs we are actually *adding* a pod not overriding like it happens for when using `:subspecs`.

The shallow clone caused the `store_pod` to remove the configuration when parsing the test spec and then the parent had 0 configuration causing an inconsistency between the parent configurations and the test spec configuration.

spec/podfile/target_definition_spec.rb
```diff
@@ -409,6 +409,15 @@ module Pod
         @child.all_whitelisted_configurations.sort.should == %w(Debug Release)
       end
 
+      it 'whitelistes pod configurations with testspecs' do
```
dnkoutso: meh: `whitelists`

PR #411, opened 2017-09-24
Add ancestor to return the outermost spec around a subspec

paulb777:
This is being added for the upcoming CocoaPods PR to fix the static_framework attribute with subspecs. See https://github.com/CocoaPods/CocoaPods/pull/6811#issuecomment-331628583

### DIFF
--- a/lib/cocoapods-core/specification/root_attribute_accessors.rb
+++ b/lib/cocoapods-core/specification/root_attribute_accessors.rb
@@ -152,6 +152,14 @@ module Pod
           command.strip_heredoc.chomp if command
         end
 
+        # @return [Hash] Return the outermost spec
+        #
+        def ancestor
+          ancestor = self
+          ancestor = ancestor.parent until ancestor.parent.nil?
+          ancestor
+        end
+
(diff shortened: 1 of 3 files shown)


lib/cocoapods-core/specification/root_attribute_accessors.rb
```diff
@@ -152,6 +152,14 @@ def prepare_command
           command.strip_heredoc.chomp if command
         end
 
+        # @return [Hash] Return the outermost spec
```
dnkoutso: is it `@return [Specification]`?

lib/cocoapods-core/specification/root_attribute_accessors.rb
```diff
@@ -152,6 +152,14 @@ def prepare_command
...
+        #
+        def ancestor
```
segiddins: This shouldn't be necessary, there's already a `Specification#root` method that returns the root specification

paulb777:
Thanks @segiddins! That's much better. Closing this.

PR #406, opened 2017-09-17
Add skip_pch podspec attribute

paulb777:
More details in associated CocoaPods PR in the next comment.

PR #412, opened 2017-09-24
Update comments/docs to indicate :skip will skip pch generation

paulb777:
Replacing #406 based on discussion at https://github.com/CocoaPods/CocoaPods/issues/6815

The corresponding CocoaPods PR is https://github.com/CocoaPods/CocoaPods/pull/7044

This PR does not change any Core functionality.

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -855,7 +855,7 @@ module Pod
       #
       #   A path to a prefix header file to inject in the prefix header of the
       #   pod project.
-      #   ':skip' indicates that the default CocoaPods prefix header should not
+      #   false indicates that the default CocoaPods prefix header should not
       #   be generated.
       #
       #   ---
@@ -867,12 +867,17 @@ module Pod
       #   @example
       #
       #     spec.prefix_header_file = 'iphone/include/prefix.pch'
-      #     spec.prefix_header_file = ':skip'
       #
-      #   @param  [String] path
-      #           The path to the prefix header file.
+      #   @example
+      #
+      #     spec.prefix_header_file = false
+      #
+      #   @param  [Bool, String] path
+      #           The path to the prefix header file or whether to disable
+      #           prefix_header generation.
       #
       attribute :prefix_header_file,
+                :types => [TrueClass, FalseClass, String],
                 :inherited => true
 
       #------------------#
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -855,15 +855,19 @@ module Pod
       #
       #   A path to a prefix header file to inject in the prefix header of the
       #   pod project.
+      #   ':skip' indicates that the default CocoaPods prefix header should not
+      #   be generated.
       #
       #   ---
       #
-      #   This attribute is __not recommended__ as Pods should not pollute the
-      #   prefix header of other libraries or of the user project.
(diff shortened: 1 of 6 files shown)


lib/cocoapods-core/specification/dsl.rb
```diff
@@ -855,15 +855,19 @@ def dependency(*args)
...
-      #   This attribute is __not recommended__ as Pods should not pollute the
-      #   prefix header of other libraries or of the user project.
+      #   The file path options is __not recommended__ as Pods should not
+      #   pollute the prefix header of other libraries or of the user project.
+      #
       #
       #   @example
       #
       #     spec.prefix_header_file = 'iphone/include/prefix.pch'
+      #     spec.prefix_header_file = ':skip'
```
segiddins: this should probably use a symbol instead of a string? Since `:skip` could very well be a valid file path

segiddins: or `spec.prefix_header_file = false` ?

paulb777: `false` is good. Made that change. Thanks!

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -867,12 +867,17 @@ def dependency(*args)
...
+      #   @example
+      #
+      #     spec.prefix_header_file = false
+      #
+      #   @param  [Bool, String] path
+      #           The path to the prefix header file or whether to disable
+      #           prefix_header generation.
       #
       attribute :prefix_header_file,
+                :types => [TrueClass, FalseClass, String],
```
segiddins: What will setting it to true do?

paulb777: I added true for symmetry with false. If it is set, it will behave the same as if the option is not set at all and the default CocoaPods pch file will be generated. 

See this test: https://github.com/paulb777/CocoaPods/blob/68abbb06a4a2d5adf9c5ce6361fab78af60be8f5/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb#L412

lib/cocoapods-core/specification/root_attribute_accessors.rb
```diff
@@ -25,6 +25,12 @@ def requires_arc
           attributes_hash['requires_arc']
         end
 
+        # @return [Bool, String] The prefix_header_file value.
+        #
+        def prefix_header_file
```
dnkoutso: this method is specified within `root_attribute_accessor` but within `dsl.rb` the declaration of of the attribute itself does not use `root_attribute` but rather `attribute`.

Is this the right place to include this method?

paulb777: I'm not sure, but I modeled it after the `requires_arc` implementation which does the same thing.

dnkoutso:
Yea probably a mistake in https://github.com/CocoaPods/Core/pull/393...

I can make another PR to clean this up, but i think `specification.rb` is where you want to include the getter.

lib/cocoapods-core/specification/dsl.rb
segiddins: 👍 but we should doc it

dnkoutso:
Gonna go ahead and merge this.

dnkoutso:
@paulb777 can you update the cocoapods PR? thank you!

paulb777:
Thanks @dnkoutso! https://github.com/CocoaPods/CocoaPods/pull/7044 is now squashed, rebased and green.

PR #417, opened 2017-10-14
[Podspec] Add `swift_version` DSL

DanToml:
Relates to: https://github.com/CocoaPods/CocoaPods/issues/7134

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -126,6 +126,21 @@ module Pod
       root_attribute :version,
                      :required => true
 
+      #------------------#
+
+      # @!method swift_version=(version)
+      #
+      #   The version of Swift that the specification supports.
+      #
+      #   @example
+      #
+      #     spec.swift_version = '3.2'
+      #
+      #   @param  [String] swift_version
+      #
+      root_attribute :swift_version,
+                     :multi_platform => false
+
       #-----------------------------------------------------------------------#
 
       # @!method cocoapods_version=(cocoapods_version)
--- a/lib/cocoapods-core/specification/root_attribute_accessors.rb
+++ b/lib/cocoapods-core/specification/root_attribute_accessors.rb
@@ -40,6 +40,14 @@ module Pod
           end
         end
 
+        # @return [Version] The swift_versions required to use the specification.
+        #
+        def swift_version
+          @swift_version ||= if version = attributes_hash['swift_version']
+                               Version.new(version)
+                             end
+        end
+
         # @return [Requirement] The CocoaPods version required to use the specification.
         #
         def cocoapods_version
--- a/spec/specification/root_attribute_accessors_spec.rb
+++ b/spec/specification/root_attribute_accessors_spec.rb
@@ -26,6 +26,14 @@ module Pod
       @spec.version.should == Version.new('1.0')
     end
 
+    describe '#swift_version' do
+      it 'return swift version of the specification' do
+        @spec.swift_version.should.be.nil
+        @spec.swift_version = '3.2'
+        @spec.swift_version.should == Version.new('3.2')
+      end
+    end
+
     it 'returns the cocoapods version requirement of the specification' do
       @spec.cocoapods_version = '>= 0.36'
       @spec.cocoapods_version.should == Requirement.new('>= 0.36')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,10 @@
 
 ##### Enhancements
 
-* None.  
+* Add `swift_version` DSL  
+  [Danielle Tomlinson](https://github.com/dantoml)
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#417](https://github.com/CocoaPods/Core/pull/417)
 
 ##### Bug Fixes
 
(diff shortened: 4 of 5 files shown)


lib/cocoapods-core/specification/root_attribute_accessors.rb
```diff
@@ -40,6 +40,15 @@ def version
           end
         end
 
+        # @return [Array<Requirement>] The swift_versions required to use the
+        #                              specification.
+        #
+        def swift_versions
+          @swift_versions ||= if versions = attributes_hash['swift_versions']
```
dnkoutso: do we need the `if versions = attributes_hash['swift_versions']`

Does it also need to be set in the `cocoapods_versions` method or does that have a default value always set?

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -126,6 +126,25 @@ module DSL
       root_attribute :version,
                      :required => true
 
+      #------------------#
+
+      # @!method swift_version=(version)
+      #
+      #   The version of Swift that the specification supports.
```
dnkoutso: Could probably expand slightly more for the end user in which we document how this is used as a "best effort" value.

lib/cocoapods-core/specification/root_attribute_accessors.rb
dnkoutso: I think `Requirement` initializer covers the case of `nil` or empty string.

dnkoutso: but i guess we would prefer a `nil` value than a `Requirement` value with `<Pod::Version version=0>`.

DanToml: I prefer a nil here for the objective-c pod case

lib/cocoapods-core/specification/dsl.rb
DanToml: Yeah, planning on following up with more info and docs

dnkoutso:
👍

orta:
Feels like good foundations 👍

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -126,6 +126,25 @@ module DSL
...
+      #     spec.swift_version = '>= 3.2'
+      #     spec.swift_versions = ['>= 3.2', '< 4.0']
+      #
+      #   @param  [String] swift_version
+      #
+      root_attribute :swift_versions,
+                     :container => Array,
+                     :singularize => true,
+                     :multi_platform => false,
+                     :root_only => true
```
dnkoutso: nit: I think `root_attribute` by default has this property set to true.

siggb:
@DanToml Hello and thanks for the improvements!

Does this enhancement resolve problem when Cocoapods override `SWIFT_VERSION` in spec dependencies? (even when I use `spec.pod_target_xcconfig  = { 'SWIFT_VERSION' => '4.0' }` command `pod spec lint` fails with error `The “Swift Language Version” (SWIFT_VERSION) build setting must be set to a supported value for targets which use Swift. This setting can be set in the build settings editor.`

siggb:
@DanToml Hello and thanks for improvements.

I've noticed a problem with spec dependencies: Cocoapods overrides `SWIFT_VERSION` on a `   Building with xcodebuild.` step.

As a result of `pod spec lint --verbose` I receive an error: `The “Swift Language Version” (SWIFT_VERSION) build setting must be set to a supported value for targets which use Swift. This setting can be set in the build settings editor.`.

dnkoutso:
@DanToml want me to rebase/merge?

lib/cocoapods-core/specification/root_attribute_accessors.rb
segiddins: nil here is the right call

lib/cocoapods-core/specification/dsl.rb
segiddins: yup

spec/specification/root_attribute_accessors_spec.rb
```diff
@@ -26,6 +26,18 @@ module Pod
...
 
+    describe '#swift_versions' do
+      it 'returns the requirements when multiple are specified' do
+        @spec.swift_versions = '>= 3.0', '< 4.0'
+        @spec.swift_versions.should == Requirement.new('>= 3.0', '< 4.0')
+      end
+
+      it 'returns the requirement when only one is specified' do
+        @spec.swift_versions = '= 3.0'
+        @spec.swift_versions.should == Requirement.new('= 3.0')
```
segiddins: nit: also test nil

DanToml:
@dnkoutso waiting till I have the main pr ready too. I don’t like merging core changes till an integration pr is at least in review

dnkoutso:
ah! ok cool sorry didn't know :)

lib/cocoapods-core/specification/dsl.rb
dnkoutso: No longer need to do this, this is just a string.

dnkoutso:
Updated per our new design decision to scale down and not support ranges. Now `swift_version` is a simple string that is wrapped by a `Version`.

This will *always* be used if the pod author had set it, otherwise we fallback to what we do today.

PR for https://github.com/CocoaPods/CocoaPods coming up.

dnkoutso:
Any objections from merging this?

dnkoutso:
Going for it!

PR #381, opened 2017-06-04
Fix Strange quotation marks in `lockfile.rb`

dacaiguoguogmail:
Strange quotation marks in 'SPEC CHECKSUMS' section
```
SPEC CHECKSUMS:
  SDWebImage: '098e97e6176540799c27e804c96653ee0833d13c'
```

PR #419, opened 2017-12-06
Use double quotes for yaml keys in Podfile.lock string representation

Ruenzuo:
This should fix an issue introduced by #381 with keys that contain !, also reported by @lcniles [here](https://github.com/CocoaPods/Core/pull/381#commitcomment-23725675).

To reproduce the problem: 
- Checkout https://github.com/Ruenzuo/reimagined-guide
- Run `bundle exec pod install`, CocoaPods will fail because Psych can't parse the Podfile.lock with keys stating with [exclamation mark](https://github.com/Ruenzuo/reimagined-guide/blob/master/Podfile.lock#L2). Error raised by Psych:

```
Psych::SyntaxError: (<unknown>): mapping values are not allowed in this context at line 48 column 28
from /Users/crisosto/.rbenv/versions/2.2.7/lib/ruby/2.2.0/psych.rb:370:in `parse'
```

I think we should keep the double quotes, but replace all single quotes to be double quotes too, otherwise we'll run again into the issue described by @orta [here](https://github.com/CocoaPods/Core/pull/381#issuecomment-306116267).

The verify the fix:
- Checkout `fixed` branch, it uses this revision.
- Delete Podfile.lock from the sample project.
- Run `bundle exec pod install` twice. It works 🎉

### DIFF
--- a/spec/lockfile_spec.rb
+++ b/spec/lockfile_spec.rb
@@ -8,13 +8,13 @@ module Pod
       <<-LOCKFILE.strip_heredoc
         PODS:
           - BananaLib (1.0):
-            - monkey (< 1.0.9, ~> 1.0.1)
+            - "monkey (< 1.0.9, ~> 1.0.1)"
           - JSONKit (1.4)
           - monkey (1.0.8)
 
         DEPENDENCIES:
           - BananaLib (~> 1.0)
-          - JSONKit (from `path/JSONKit.podspec`)
+          - "JSONKit (from `path/JSONKit.podspec`)"
 
         EXTERNAL SOURCES:
           JSONKit:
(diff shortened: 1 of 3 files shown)


spec/lockfile_spec.rb
```diff
@@ -8,13 +8,13 @@ def self.yaml
       <<-LOCKFILE.strip_heredoc
         PODS:
           - BananaLib (1.0):
-            - monkey (< 1.0.9, ~> 1.0.1)
+            - "monkey (< 1.0.9, ~> 1.0.1)"
```
segiddins: these strings shouldn't need to be quoted?

Ruenzuo: I think this is something `Pod::YAMLHelper` is doing, but the behaviour was previously hidden by the code that removed all quote marks. When we process string values we [escape and quote special characters with `#inspect`](https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/yaml_helper.rb#L293-L303) but those quotes are removed afterwards by [the code this PR is changing](https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/lockfile.rb#L354-L355). 

What should be the right approach here?

PR #421, opened 2018-01-22
[Lockfile] Store what spec repo a pod comes from

segiddins:
For https://github.com/CocoaPods/CocoaPods/pull/7358

### DIFF
--- a/lib/cocoapods-core/lockfile.rb
+++ b/lib/cocoapods-core/lockfile.rb
@@ -121,6 +121,17 @@ module Pod
       @dependencies
     end
 
+    # Returns pod names grouped by the spec repo they were sourced from.
+    #
+    # @return [Hash<String, Array<String>>] A hash, where the keys are spec
+    #         repo source URLs (or names), and the values are arrays of pod names.
+    #
+    # @note   It does not include pods that come from "external sources".
+    #
+    def pods_by_spec_repo
+      @pods_by_spec_repo ||= internal_data['SPEC REPOS']
+    end
+
     # Generates a dependency that requires the exact version of the Pod with the
     # given name.
     #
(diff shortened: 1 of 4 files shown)


lib/cocoapods-core/lockfile.rb
```diff
@@ -440,6 +453,13 @@ def generate_dependencies_data(podfile)
         podfile.dependencies.map(&:to_s).sort
       end
 
+      def generate_spec_repos(spec_repos)
```
dnkoutso: nit if you care: add doc for this method

lib/cocoapods-core/lockfile.rb
```diff
@@ -440,6 +453,13 @@ def generate_dependencies_data(podfile)
...
+        spec_repos.map do |source, specs|
+          next unless source
+          [source.url || source.name, specs.map(&:name)]
+        end.compact.to_h
```
segiddins: `to_h` doesn't exist on 2.0.0

dnkoutso: <3

lib/cocoapods-core/lockfile.rb
segiddins: should this filter out sources with no specs? probably

lib/cocoapods-core/yaml_helper.rb
```diff
@@ -143,6 +145,8 @@ def process_array(array)
       # @return [String] the YAML representation of the given object.
       #
       def process_hash(hash, hash_keys_hint = nil, line_separator = "\n")
+        return '{}' if hash.empty?
```
segiddins: this breaks "empty" lockfiles, so we might need to add a test for empty lockfiles explicitly?

segiddins: no it doesn't, since we always add the CocoaPods version. The change in the integration spec is expected (and correct), since we're manually removing the cocoapods version (to avoid noisy diffs)

segiddins:
Feedback addressed

dnkoutso:
💯

PR #422, opened 2018-01-22
[Dependency] Have #merge take into account podspec repos

segiddins:


dnkoutso:
missing `CHANGELOG` entry but 👍

spec/dependency_spec.rb
```diff
@@ -224,6 +237,42 @@ module Pod
...
+      it 'allows merging if both dependencies have the same podspec repo' do
+        dep1 = Dependency.new('bananas', '~> 1.9', :source => 'https://source.git')
+        dep2 = Dependency.new('bananas', :source => 'https://source.git')
+
+        expected = Dependency.new('bananas', '~> 1.9', :source => 'https://source.git')
+        dep1.merge(dep2).should == expected
+        dep2.merge(dep1).should == expected
+      end
+
+      it 'raises when attempting to merge with a podspec repo and external source' do
```
dnkoutso: is this the same test as above?

dnkoutso: ah nevermind its external repo

lib/cocoapods-core/dependency.rb
```diff
@@ -95,8 +95,7 @@ def initialize(name = nil, *requirements)
             raise Informative, 'A dependency with a specified podspec repo may ' \
               "not include other source parameters (#{name})."
           end
-        else
-          @external_source = additional_params
+        elsif @external_source = additional_params
           unless requirements.empty?
             raise Informative, 'A dependency with an external source may not ' \
```
justinseanmartin: I don't understand when/why we would expect dependencies with an external source to not specify a version requirement.

lib/cocoapods-core/dependency.rb
```diff
@@ -249,24 +249,41 @@ def merge(other)
...
             end
 
+      opts = {}
+
       if external_source || other.external_source
-        self_external_source  = external_source || {}
-        other_external_source = other.external_source || {}
-        dep.external_source = self_external_source.merge(other_external_source)
+        opts.
+          merge!(external_source || {}) { raise ArguementError, 'key conflict' }.
```
justinseanmartin: This exception isn't helpful, should update to include which key conflicts and what their values are.

dnkoutso: Also `ArgumentError`

lib/cocoapods-core/dependency.rb
segiddins: agreed, but I dont think this is the place to change that

dnkoutso:
same needs a rebase

segiddins:
Feedback addressed

PR #423, opened 2018-01-22
Unset GIT_CONFIG when running commands

keith:
GIT_CONFIG is an environment variable that overrides which files `config
--get` actually fetches from. The benefit in setting this is that if you
want to add configuration, you don't have to use `--global`. I would
never recommend anyone set this, but in the case that you do, CocoaPods
can never find local specs repos by matching URLs, since getting
remote.origin.url will come from a different file, and likely always be
empty.

There's a `--local` flag that is also supposed to be the opposite of
`--global` for this, but setting this when GIT_CONFIG is set results in
an error. The other option would be to pass the file explicitly with
`--file .git/config`, but this feels more fragile.

### DIFF
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -435,7 +435,7 @@ module Pod
     end
 
     def repo_git(args, include_error: false)
-      command = "git -C \"#{repo}\" " << args.join(' ')
+      command = "env -u GIT_CONFIG git -C \"#{repo}\" " << args.join(' ')
       command << ' 2>&1' if include_error
       (`#{command}` || '').strip
     end
--- a/spec/master_source_spec.rb
+++ b/spec/master_source_spec.rb
@@ -25,8 +25,8 @@ module Pod
       end
 
       it 'uses the only fast forward git option' do
-        @source.expects(:`).with("git -C \"#{@path}\" checkout master")
-        @source.expects(:`).with("git -C \"#{@path}\" pull --ff-only 2>&1")
+        @source.expects(:`).with("env -u GIT_CONFIG git -C \"#{@path}\" checkout master")
+        @source.expects(:`).with("env -u GIT_CONFIG git -C \"#{@path}\" pull --ff-only 2>&1")
         @source.send :update_git_repo
       end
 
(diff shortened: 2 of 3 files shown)


keith:
Here's a similar issue on homebrew https://github.com/Homebrew/brew/issues/160

dnkoutso:
Does this apply only to the `repos` related commands or everything? Is it intentional to scope it only for `repo_git` command?

keith:
Anything running git commands would have this issue

keith:
Sorry, this really only affects `git config` related commands I guess. I can try to scope it to that instead if that's what you mean?

keith:
Updated with a different approach, the problem was setting GIT_CONFIG to empty was not the same as unsetting it from git's perspective. I also fixed the tests.

keith:
If we generally don't want this I'm fine with:

1. Not merging this and commenting on the issue that comes up when you google the resulting error
2. Just throwing an Informative exception from somewhere here if you have that env var set

segiddins:
I think I'd prefer either of those options?

What actually breaks from trying to use that env var?

keith:
You can’t pod install since it fails to find a repo with a url matching the source defined in the Podfile.

segiddins:
> You can’t pod install since it fails to find a repo with a url matching the source defined in the Podfile.

I think that's only if you're pointing to a config that overwrites remote/branch info? If I only use the file to override something in `[user]` I don't think it breaks anything? I might be mistaken, though, since this is the first time I've seen that env var.

keith:
Unfortunately it breaks everything. Setting this variable, even to an empty string causes git to search for `remote.origin.url` in that file. `pod install` ends up attempting to re-setup the master repo (or your custom repo) every single time you run it

keith:
It doesn't intelligently fallback to any other configuration when this is set

segiddins:
Ah gotcha. Yeah, I see running `git config` no other settings are there, but for some reason `git fetch --verbose` still fetches from the correct remote. In that case, I'm 👍 on this

dnkoutso:
needs a rebase and we can merge!

keith:
Rebased!

PR #425, opened 2018-01-25
Add Range header to HTTP::perform_head_request

### LINKED ISSUE #420: Add Range header to GET request when used as a fallback in perform_head_request
I was wondering if it would make sense to add a `Range` header to the [GET request in `HTTP.perform_head_request`](https://github.com/CocoaPods/Core/blob/33299ada83b98e1c7ec117f4d4d67055b2247a19/lib/cocoapods-core/http.rb#L68).

# Background
The GET request is being used as a fallback if the HEAD request returns a failure code, since some hosts have inconsistencies between HEAD and GET (for example, returning 200 OK for GET and 403 Forbidden for HEAD).

[_trunk.cocoapods.org_](https://github.com/CocoaPods/trunk.cocoapods.org) is using the [`HTTP.perform_head_request`](https://github.com/CocoaPods/Core/blob/33299ada83b98e1c7ec117f4d4d67055b2247a19/lib/cocoapods-core/http.rb#L62) when it calls out to [`HTTP.validate_url`](https://github.com/CocoaPods/Core/blob/33299ada83b98e1c7ec117f4d4d67055b2247a19/lib/cocoapods-core/http.rb#L41) in the check for [`SpecificationWrapper.validate_http`](https://github.com/CocoaPods/trunk.cocoapods.org/blob/a9fe3f5f74cb97927827796217da66c5d387c3d9/app/models/specification_wrapper.rb#L73). The `SpecificationWrapper.validate_http` also includes a timeout wrapper for 5 seconds.

# Problem description
If both the HEAD request fails and the requested file is large enough to take longer than 5 seconds, then the file http is not considered valid, even though under normal circumstances it would be.

# Proposed solution
One solution to this would be to add a [`Range`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) header to the GET request so that the server only returns a small portion of the requested file, thus not tripping the timeout wrapper that _trunk.cocoapods.org_ is using.

# Possible issues
Some servers won't support Range requests, but the use of GET in this case is to test for the existences, so we don't need the whole file and we don't need the ability to resume downloading (my impression is that this would normally be the main reason for using a Range request). If Range is not supported, the server will likely either ignore the header and return the response it otherwise would have sent, or the server refuses to proceed. The edge case here would be that the server returns a failure code because of the Range request even though it might have otherwise returned a success code. In this case there might need to be a third fallback that doesn't include the Range request. I'm not sure how common this would be.

# Example code
An example of using the header would be the following (I'm using the files mentioned in [this bug report](https://github.com/CocoaPods/trunk.cocoapods.org/issues/236) for the example):

### HEAD request fails:
```
$ curl --location --head https://github.com/kylefleming/opencv/releases/download/3.3.0/opencv-3.3.0-ios-osx-framework.zip
...
HTTP/1.1 403 Forbidden
x-amz-request-id: C90A15FD69DC2813
x-amz-id-2: 6DORgM622YIDftqj4ttJIx5qnuCU0Rrj3fAKqSLKU40jKYC8hFq/ejVPF/HUjl+O
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Fri, 12 Jan 2018 03:05:13 GMT
Server: AmazonS3
```

### GET request takes more than 5 seconds (22 seconds on my machine):
```
$ curl --location https://github.com/kylefleming/opencv/releases/download/3.3.0/opencv-3.3.0-ios-osx-framework.zip > opencv-3.3.0-ios-osx-framework.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   624    0   624    0     0    624      0 --:--:-- --:--:-- --:--:--  2526
100  231M  100  231M    0     0  10.5M      0  0:00:22  0:00:22 --:--:-- 9418k
```

### The proposed added step of using a GET request with a Range header (this example requests the first byte of the content):
```
$ curl --location --header "Range: bytes=0-0" https://github.com/kylefleming/opencv/releases/download/3.3.0/opencv-3.3.0-ios-osx-framework.zip > opencv-3.3.0-ios-osx-framework.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   624    0   624    0     0    624      0 --:--:-- --:--:-- --:--:--  2678
100     1  100     1    0     0      1      0  0:00:01 --:--:--  0:00:01  1000
```

# Summary:
The ultimate goal here would be to allow for a GET fallback that more easily supports larger files when calling [`HTTP.perform_head_request`](https://github.com/CocoaPods/Core/blob/33299ada83b98e1c7ec117f4d4d67055b2247a19/lib/cocoapods-core/http.rb#L62) by not requesting the entire file.

kylefleming:
Uses a GET request with a 1 byte Range header if the initial HEAD request fails. Falls back to a GET request without a Range header if that also fails.

Closes #420

### Discussion
I've added the GET request with a Range header as an additional fallback and kept the original non-Range GET request as the final fallback. This is because I wasn't sure if we should expect that a server might respond to a normal GET request but return an error if the GET request includes a Range header.

My intuition would say that almost all servers that don't respond to a Range header would just ignore it and proceed as if it didn't exist. In this case, we wouldn't need the third non-Range fallback. Do we think this is a case that we would want to support? Having all 3 is the more conservative route, but adds additional latency in returning a result if the server doesn't have the content at all.

I think ideally we'd remove the third non-Range fallback, but I wanted to submit this PR with it in, in case it is decided that it's necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ##### Enhancements
 
+* `HTTP::perform_head_request` now includes a 1-byte `Range` header in the fallback GET
+  request.
+  [Kyle Fleming](https://github.com/kylefleming)
+  [#425](https://github.com/CocoaPods/Core/pull/425)
+
 * Update Podfile Reference Guide to include `:source` parameter for the `pod` statement  
   [Mark Woollard](https://github.com/mwoollard)
   [#7359](https://github.com/CocoaPods/CocoaPods/issues/7359)
--- a/lib/cocoapods-core/http.rb
+++ b/lib/cocoapods-core/http.rb
@@ -44,7 +44,7 @@ module Pod
       begin
         url = get_actual_url(url)
         resp = perform_head_request(url)
-      rescue SocketError, URI::InvalidURIError, REST::Error, REST::DisconnectedError
+      rescue SocketError, URI::InvalidURIError, REST::Error, REST::Error::Connection
         resp = nil
       end
 
@@ -65,7 +65,12 @@ module Pod
       resp = ::REST.head(url, 'User-Agent' => USER_AGENT)
 
       if resp.status_code >= 400
-        resp = ::REST.get(url, 'User-Agent' => USER_AGENT)
+        resp = ::REST.get(url, 'User-Agent' => USER_AGENT,
+                               'Range' => 'bytes=0-0')
+
+        if resp.status_code >= 400
+          resp = ::REST.get(url, 'User-Agent' => USER_AGENT)
+        end
       end
 
       resp
--- a/spec/http_spec.rb
+++ b/spec/http_spec.rb
@@ -2,6 +2,10 @@ require File.expand_path('../spec_helper', __FILE__)
 
 module Pod
   describe HTTP do
+    after do
+      WebMock.reset!
+    end
+
     describe 'In general' do
       it 'can resolve redirects' do
         WebMock::API.stub_request(:head, /redirect/).to_return(
@@ -28,7 +32,7 @@ module Pod
         response.success?.should.be.true
       end
 
-      it 'is resilient to HEAD errros' do
+      it 'is resilient to HEAD errors' do
         WebMock::API.stub_request(:head, /foo/).to_return(:status => 404)
         WebMock::API.stub_request(:get, /foo/).to_return(:status => 200)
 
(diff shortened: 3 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -45,7 +45,7 @@ def copy_fixture_to_pod(name, pod)
 VCR.configure do |c|
   c.cassette_library_dir = ROOT + 'spec/fixtures/vcr_cassettes'
   c.hook_into :webmock
-  c.ignore_hosts 'codeclimate.com'
+  c.ignore_hosts 'codeclimate.com', 'some-url'
```
segiddins: I dont think this host should be ignored?

kylefleming: You're right.

I was thinking that if you forget to use `WebMock` to mock an endpoint, then `VCR` shouldn't try to handle it since it's not a legitimate URL to begin with and will never be recorded as a cassette. However, it's better that `VCR` complains about it than for `Net::HTTP` to try to connect to it.

kylefleming:
@segiddins Do you think it's preferable to have the original fallback GET request as a third fallback case, or just get rid of it entirely?

segiddins:
> Do you think it's preferable to have the original fallback GET request as a third fallback case, or just get rid of it entirely?

I think its a preferable fallback

kylefleming:
Shall I squash it?

dnkoutso:
@kylefleming sure

dnkoutso:
danke!

kylefleming:
Do we have an estimate on when 1.5 will be released? Still waiting on this fix so I can push the latest OpenCV pod (CocoaPods/trunk.cocoapods.org#236).

segiddins:
@kylefleming not yet, but you can point to Core master using a gemfile to use it now: https://guides.cocoapods.org/using/unreleased-features.html

kylefleming:
@segiddins Thanks for the suggestion, but I actually need https://github.com/CocoaPods/trunk.cocoapods.org to be updated with this fix in order to push the pod.

kylefleming:
@segiddins Do you think it will be awhile before it's released? I'm inquiring so that I know whether I should be looking into workarounds or not.

segiddins:
I don't know how long it will be

PR #427, opened 2018-01-31
[Set] Speed up perfomance

segiddins:


lib/cocoapods-core/specification/set.rb
```diff
@@ -68,7 +68,7 @@ def specification_paths_for_version(version)
       #         from highest to lowest.
       #
       def versions
-        versions_by_source.values.flatten.uniq.sort.reverse
+        @versions ||= versions_by_source.values.flatten.uniq.sort.reverse
```
DanToml: This could be invalidated by a core user who got versions, caused a spec update, and gets versions again

DanToml: actually nvm we already cache versions_by_source, this should be fine.

segiddins: Dang. Can you think of any way around that?

PR #435, opened 2018-03-01
Fix crash when there's an empty source spec directory

paulb777:
Fix https://github.com/CocoaPods/CocoaPods/issues/6381

If a source spec directory has had all of its version subdirectories removed, `pod install` would crash with `TypeError - no implicit conversion of nil into String` at `cocoapods-core/specification.rb:584`.

The empty directory would occur if a Spec repo previously had one or more versions of a particular pod and then all of the versions were removed.

The workaround was to delete ~/.cocoapods/repos/{repo}/ or just the offending empty directory.

### DIFF
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -262,7 +262,13 @@ module Pod
       if query.is_a?(Dependency)
         query = query.root_name
       end
-      found = Pathname.glob(pod_path(query)).map { |path| path.basename.to_s }
+
+      found = []
+      Pathname.glob(pod_path(query)) do |path|
+        next unless Dir.foreach(path).any? { |child| child != '.' && child != '..' }
+        found << path.basename.to_s
+      end
+
       if [query] == found
         set = set(query)
         set if set.specification_name == query
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
   
 ##### Bug Fixes
 
+* Fix crash when there's an empty source spec directory  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#6381](https://github.com/CocoaPods/CocoaPods/issues/6381)
+
 * The `Dependency#merge` method takes into account any `podspec_repo`s the dependencies
   may have set.  
   [Samuel Giddins](https://github.com/segiddins)
--- a/spec/source_spec.rb
+++ b/spec/source_spec.rb
@@ -131,6 +131,15 @@ module Pod
         @source.search('bAnAnAlIb').should.be.nil?
       end
 
+      describe 'when there is an empty directory' do
+        before { @empty_dir = @path.join('Specs', 'Empty').tap(&:mkpath) }
+        after { FileUtils.rm_r @empty_dir }
+
+        it 'returns nil' do
+          @source.search('Empty').should.be.nil
+        end
+      end
+
       describe '#search_by_name' do
         it 'properly configures the sources of a set in search by name' do
           source = Source.new(fixture('spec-repos/test_repo'))


dnkoutso:
👍 with a test

lib/cocoapods-core/source.rb
```diff
@@ -262,7 +262,7 @@ def search(query)
       if query.is_a?(Dependency)
         query = query.root_name
       end
-      found = Pathname.glob(pod_path(query)).map { |path| path.basename.to_s }
+      found = Pathname.glob(pod_path(query)).map { |path| Dir[path.to_s + '/*'].empty? ? nil : path.basename.to_s }
```
dnkoutso: do we care if `nil` values end up in the array? Do we want to add `.compact` in the end here?

segiddins:
I'm nervous about us having to go back to the FS so much -- this looks like it might be suboptimal for performance

segiddins:
I'd prefer something like this, I think:

```diff
diff --git a/lib/cocoapods-core/source.rb b/lib/cocoapods-core/source.rb
index 47375c46..8e84afd7 100644
--- a/lib/cocoapods-core/source.rb
+++ b/lib/cocoapods-core/source.rb
@@ -262,7 +262,13 @@ module Pod
       if query.is_a?(Dependency)
         query = query.root_name
       end
-      found = Pathname.glob(pod_path(query)).map { |path| path.basename.to_s }
+
+      found = []
+      Pathname.glob(pod_path(query)) do |path|
+        next if path.empty?
+        found << path.basename.to_s
+      end
+
       if [query] == found
         set = set(query)
         set if set.specification_name == query
diff --git a/spec/source_spec.rb b/spec/source_spec.rb
index 5ef2e088..5a7f65c2 100644
--- a/spec/source_spec.rb
+++ b/spec/source_spec.rb
@@ -131,6 +131,15 @@ module Pod
         @source.search('bAnAnAlIb').should.be.nil?
       end
 
+      describe "when there is an empty directory" do
+        before { @empty_dir = @path.join('Specs', 'Empty').tap(&:mkpath) }
+        after { FileUtils.rm_r @empty_dir }
+
+        it 'returns nil' do
+          @source.search('Empty').should.be.nil
+        end
+      end
+
       describe '#search_by_name' do
         it 'properly configures the sources of a set in search by name' do
           source = Source.new(fixture('spec-repos/test_repo'))

```

paulb777:
Redone with @segiddins recommended code. Thanks!

paulb777:
Looks like empty? does not exist before ruby 2.4

lib/cocoapods-core/source.rb
```diff
@@ -262,7 +262,13 @@ def search(query)
       if query.is_a?(Dependency)
         query = query.root_name
       end
-      found = Pathname.glob(pod_path(query)).map { |path| path.basename.to_s }
+
+      found = []
+      Pathname.glob(pod_path(query)) do |path|
+        next if Dir[path.to_s + '/*'].empty?
```
dnkoutso: Is `Dir[something].empty?` what @segiddins was trying to avoid? Does this hit the FS? I think in his commit this was not using `Dir`

paulb777: Is there an alternative that works on pre-2.4 ruby's or is there a way to run different code depending on ruby versions? See the stackoverflow question referenced above.

dnkoutso: ah sorry did not connect the two, yeah I am uncertain we can wait for @segiddins to weigh in.

segiddins: `next if Dir(path).empty?`

segiddins: ugh looks like thats 2.4+ as well.

`next unless Dir.each_child(path).any?` might work? (the docs dont say when each method was introduced)

paulb777: each_child was introduced in 2.5 - https://blog.bigbinary.com/2017/11/21/ruby-2_5-introduces-dir-children-and-dir-each_child.html

segiddins: ugh such terrible docs

maybe `Dir.foreach(path).any? {|child| child != '.' && child != '.. }`

I just really want to avoid globbing whenever possible, as it's expensive

PR #443, opened 2018-04-10
[Lockfile] Improve performance of #detect_changes_with_lockfile

segiddins:
By using a Hash instead of Array#find, this becomes O(n) rather than O(n^2)

Additionally, we no longer need to call Podfile#dependencies once per dependency name, which is huge since that list of dependencies gets recomputed on each call


dnkoutso:
Danger still broken here so you won't get a green checkmark, merge away!

PR #444, opened 2018-04-11
Provide a helpful error message when attempting to assign a value to `dependency`

amorde:
closes https://github.com/CocoaPods/CocoaPods/issues/7576

I've definitely made the same mistake before.

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -666,6 +666,12 @@ module Pod
         attributes_hash['dependencies'][name] = version_requirements
       end
 
+      def dependency=(args)
+        joined = args.join('\', \'')
+        arguments = "\'#{joined}\'"
+        raise Informative, "Cannot assign value to `dependency`. Did you mean: `dependency #{arguments}`?"
+      end
+
       #------------------#
 
       # @!method requires_arc=(flag)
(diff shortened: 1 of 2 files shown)


lib/cocoapods-core/specification/dsl.rb
```diff
@@ -666,6 +666,12 @@ def dependency(*args)
         attributes_hash['dependencies'][name] = version_requirements
       end
 
+      def dependency=(args)
+        joined = args.join('\', \'')
+        arguments = "\'#{joined}\'"
+        raise Informative, "Cannot assign value to `dependency`. Did you mean: `dependency #{arguments}`?"
```
segiddins: thing we could just do `dependency #{args.map(&:inspect).join(', ')`

dnkoutso:
Just needs a rebase

lib/cocoapods-core/specification/dsl.rb
amorde: oh cool. didn't know that existed

amorde: this results in escaped quotes: `dependency \"JSONKit\", \"1.5\"`

dnkoutso:
ping @amorde if you get a chance to rebase so we can land this.

amorde:
Rebased

segiddins:
Awesome!

amorde:
Will this show up on the Podspec syntax guide? I'd hope not, but I can imagine it would

segiddins:
It shouldn't, since it doesn't have any docs

PR #445, opened 2018-04-17
[Specification] Update resource_bundles documentation for usage recommendation ambiguity 🌈

yoonapps:
As brought up in [#5192](https://github.com/CocoaPods/CocoaPods/issues/5192), the documentation for preferring the use of `resource_bundles` over `resources` is outdated with the integration of frameworks. Seemed like @orta and @neonichu agreed on updating the docs so here's my crack at it.

tl;dr I specified that using `resource_bundles` is preferred for Pods that are built as static libraries.

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1157,9 +1157,9 @@ module Pod
       #   hash where the keys represent the name of the bundles and the values
       #   the file patterns that they should include.
       #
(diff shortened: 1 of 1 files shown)


yoonapps:
I wasn't sure I should add that `resources` is preferred (is it?) for dynamic frameworks. I can definitely update my PR. All feedback is welcome!

orta:
Yeah, looks good to me - thanks

PR #455, opened 2018-06-21
[Version] Stop parsing legacy HEAD versions

segiddins:
It's been 2 years since they were removed, and this saves a regexp match on every version allocation


dnkoutso:
@segiddins rebased it for you.

PR #456, opened 2018-08-08
Use the `podspec :subspec=>'xx'` to load dependencies from a subspec

Whirlwind:
Sometimes, I use the `podspec` in a Podfile to load dependencies from a `.podspec` file. But I often just load one or some subspec(s) from the podspec file, not load all dependencies.
eg: this is a Podspec:
```
s.dependency 'SDWebImage'
s.subspec 'ASI' do |s1|
    s1.dependency 'ASIHTTPRequest'
end
s.subspec 'AF' do |s2|
    s2.dependency 'AFNetworking'
end
```
this is a Podfile:
```
target :SDKForASI do
    podspec :subspec=>'ASI'
end
target :SDKForAF do
    podspec :subspec=>'AF'
end
```

I need the target have different dependencies from different subspecs. This PR will add a option `subspec`(`subspecs`) to the `podspec` DSL.

### DIFF
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -663,15 +663,24 @@ module Pod
       # @return [void]
       #
       def store_podspec(options = nil)
-        if options
-          unless options.keys.all? { |key| [:name, :path].include?(key) }
-            raise StandardError, 'Unrecognized options for the podspec ' \
-              "method `#{options}`"
-          end
-          get_hash_value('podspecs', []) << options
-        else
-          get_hash_value('podspecs', []) << { :autodetect => true }
+        options ||= {}
+        unless options.keys.all? { |key| [:name, :path, :subspecs, :subspec].include?(key) }
+          raise StandardError, 'Unrecognized options for the podspec ' \
+            "method `#{options}`"
         end
+        if subspec_name = options[:subspec]
+          unless subspec_name.is_a?(String)
+            raise StandardError, "Option `:subspec => #{subspec_name.inspect}` should be a String"
+          end
+        end
+        if subspec_names = options[:subspecs]
+          if !subspec_names.is_a?(Array) || !subspec_names.all? { |name| name.is_a? String }
+            raise StandardError, "Option `:subspecs => #{subspec_names.inspect}` " \
+            'should be an Array of Strings'
+          end
+        end
+        options[:autodetect] = true if !options.include?(:name) && !options.include?(:path)
+        get_hash_value('podspecs', []) << options
       end
 
       #--------------------------------------#
@@ -907,9 +916,18 @@ module Pod
         podspecs.map do |options|
           file = podspec_path_from_options(options)
           spec = Specification.from_file(file)
-          all_specs = [spec, *spec.recursive_subspecs]
-          all_deps = all_specs.map { |s| s.dependencies(platform) }.flatten
-          all_deps.reject { |dep| dep.root_name == spec.root.name }
+          subspec_names = options[:subspecs] || options[:subspec]
+          specs = if subspec_names.blank?
+                    [spec]
+                  else
+                    subspec_names = [subspec_names] if subspec_names.is_a?(String)
+                    subspec_names.map { |subspec_name| spec.subspec_by_name("#{spec.name}/#{subspec_name}") }
+                  end
+          specs.map do |subspec|
+            all_specs = [subspec, *subspec.recursive_subspecs]
+            all_deps = all_specs.map { |s| s.dependencies(platform) }.flatten
+            all_deps.reject { |dep| dep.root_name == subspec.root.name }
+          end.flatten
         end.flatten.uniq
       end
 
--- a/spec/podfile/dsl_spec.rb
+++ b/spec/podfile/dsl_spec.rb
@@ -405,7 +405,16 @@ module Pod
           platform :ios
           podspec :path => banalib_path
         end
-        podfile.dependencies.map(&:name).should == %w(monkey)
+        podfile.dependencies.map(&:name).should == %w(monkey AFNetworking SDWebImage)
+      end
+
+      it 'it can use use the dependencies of a podspec\'s subspec' do
+        banalib_path = fixture('BananaLib.podspec').to_s
+        podfile = Podfile.new(fixture('Podfile')) do
+          platform :ios
+          podspec :path => banalib_path, :subspec => 'GreenBanana'
+        end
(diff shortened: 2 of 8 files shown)


amorde:
I don't understand the issue this is trying to fix. How is this different from just specifying the subspec directly? Like this:

```
target :SDKForASI do
    pod 'MySDK/ASI'
end
target :SDKForAF do
    pod 'MySDK/AF'
end
```

Whirlwind:
Maybe I lost some description. The project and targets is the development pods.
The `MySDK/ASI`'s `source_files` will be installed in `Pods` project and build target `Pods-MySDK-ASI.a`. But my target `SDKForASI` includes the `source_files` and has been compiled by xcode. So If I use `pod 'MySDK/ASI'`, I will get duplicate symbols and classes.

If I use `podspec`, CocoaPods will only pick the dependencies from  podspec file, will not install the `MySDK`.

amorde:
Ah, okay. So you're using this syntax https://guides.cocoapods.org/syntax/podfile.html#podspec to load the dependencies of the podspec.

Is the issue that you want only the dependencies of the `subspec`, and _not_ the dependencies of the root spec? Does `podspec 'MySDK/ASI'` not support what you're trying to do?

Whirlwind:
The `name` is used to search the podspec file, it could not pick the dependencies from subspec.

amorde:
Ok thanks, I think I understand now

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -663,15 +663,13 @@ def store_pod(name, *requirements)
...
-        if options
-          unless options.keys.all? { |key| [:name, :path].include?(key) }
-            raise StandardError, 'Unrecognized options for the podspec ' \
-              "method `#{options}`"
-          end
-          get_hash_value('podspecs', []) << options
-        else
-          get_hash_value('podspecs', []) << { :autodetect => true }
+        options ||= {}
+        unless options.keys.all? { |key| [:name, :path, :subspecs, :subspec].include?(key) }
```
amorde: I think its worth adding only the plural `:subspecs` option to keep the implementation simpler. If you only want 1 subspec, that could just be `:subspecs => ['Subspec']`

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -907,9 +905,18 @@ def podspec_dependencies
         podspecs.map do |options|
           file = podspec_path_from_options(options)
           spec = Specification.from_file(file)
-          all_specs = [spec, *spec.recursive_subspecs]
-          all_deps = all_specs.map { |s| s.dependencies(platform) }.flatten
-          all_deps.reject { |dep| dep.root_name == spec.root.name }
+          subspec_names = options[:subspecs] || options[:subspec]
+          specs = if subspec_names.blank?
```
amorde: Could this raise an error if the type is incorrect? Like if we receive `:subspecs => [1234]`

spec/podfile/dsl_spec.rb
```diff
@@ -405,7 +405,16 @@ module Pod
           platform :ios
           podspec :path => banalib_path
         end
-        podfile.dependencies.map(&:name).should == %w(monkey)
+        podfile.dependencies.map(&:name).should == %w(monkey AFNetworking SDWebImage)
+      end
+
+      it 'it can use use the dependencies of a podspec\'s subspec' do
```
amorde: nit: remove the extra 'it'

lib/cocoapods-core/podfile/target_definition.rb
Whirlwind: I think about it. But I found that I often only use one subspec in a target, so the `subspec=>'xx'` is very simple and very common.

lib/cocoapods-core/podfile/target_definition.rb
Whirlwind: Is it strange that each parameter determines the type?

spec/podfile/dsl_spec.rb
Whirlwind: 😂I just copy the description from last spec ( Line 402 ). Should I remove them?

lib/cocoapods-core/podfile/target_definition.rb
amorde: Well they sort of have a type already, even if its implicit rather than explicit. Passing a `Hash` or something else into `:subspecs` doesn't really make sense

spec/podfile/dsl_spec.rb
amorde: just the the beginning so it's `it 'can use use the dependencies of a podspec\'s subspec' do`

lib/cocoapods-core/podfile/target_definition.rb
Whirlwind: add `subspec_names.map!(&:to_s)` ?

Whirlwind: It seams that it will not raise an error if the `subspecs => [1234]`.
```
          subspec_names = options[:subspecs] || options[:subspec]
          specs = if subspec_names.blank?
                    [spec]
                  else
                    subspec_names = [subspec_names] if subspec_names.is_a?(String)
                    subspec_names.map { |subspec_name| spec.subspec_by_name("#{spec.name}/#{subspec_name}") }
                  end
```
It will raise an error if `:subspec=>123`.

Whirlwind:
@amorde What description should I add into the changelog?

lib/cocoapods-core/podfile/target_definition.rb
amorde: Yeah, I'd rather not just convert everything to a String since that becomes an expected functionality of the API

It makes sense that if you are specifying a subspec name, it should be a String.

amorde:
Something like

```
Add support for specifying subspecs when using the `podspec` Podfile attribute
```

lib/cocoapods-core/podfile/target_definition.rb
Whirlwind: I add the type checking in the `store_podspec` method.

Whirlwind:
There are always some strange questions recently in CI. I pass all unit tests in my mac, and pass them under some ruby versions but other version.

amorde:
@Whirlwind I think this is good to go! Could you squash the commits?

dnkoutso:
+1 on squashing commits

Whirlwind:
@amorde @dnkoutso done. Thanks.

amorde:
Thanks @Whirlwind!

PR #457, opened 2018-08-10
Allow duplicate dependencies when they have no conflict

Whirlwind:
see https://github.com/CocoaPods/CocoaPods/pull/7999


Whirlwind:
The Ruby 2.3.4 seems that have some problem.

dnkoutso:
@Whirlwind want to rebase and update this PR?

Whirlwind:
@dnkoutso Do you agree with my idea? If yes, I will update this PR.

dnkoutso:
@Whirlwind I do not think this will be implemented currently. I am going to close the PR as it has been lingering for a while. I suggest we also close #7999.

PR #460, opened 2018-09-04
Add `type` attribute to `Specification` and use that to determine if it's a test_spec

dostrander:
🌈 🌈 🌈 🌈 🌈 🌈
Doesn't change any public API however internals of the `Specification` now uses a `.type` attribute to determine if it's a `test_specification`. 

This aims to keep the API the same while still allowing for extensibility of top level specs. A 

@dnkoutso @segiddins 
 🌈 🌈 🌈 🌈 🌈 🌈

### DIFF
--- a/lib/cocoapods-core/specification.rb
+++ b/lib/cocoapods-core/specification.rb
@@ -34,18 +34,18 @@ module Pod
     # @param  [String] name
     #         the name of the specification.
     #
-    # @param [Bool] test_specification
-    #        Whether the specification is a test specification
+    # @param [Symbol] type
+    #        the type of specification.
     #
-    def initialize(parent = nil, name = nil, test_specification = false)
+    def initialize(parent = nil, name = nil, type = :root)
       @attributes_hash = {}
       @subspecs = []
       @consumers = {}
       @parent = parent
       @hash_value = nil
-      @test_specification = test_specification
       attributes_hash['name'] = name
-      attributes_hash['test_type'] = :unit if test_specification
+      attributes_hash['type'] = type
+      attributes_hash['test_type'] = :unit if type == :test
 
       yield self if block_given?
     end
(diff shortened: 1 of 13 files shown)


dostrander:
@segiddins @dnkoutso

lib/cocoapods-core/specification.rb
```diff
@@ -34,18 +34,18 @@ class Specification
     # @param  [String] name
     #         the name of the specification.
     #
-    # @param [Bool] test_specification
-    #        Whether the specification is a test specification
+    # @param [Symbol] type
+    #        the type of specification.
     #
-    def initialize(parent = nil, name = nil, test_specification = false)
+    def initialize(parent = nil, name = nil, type = :root)
```
segiddins: This is probably a breaking change for anything that used `Pod::Specification.new` ?

lib/cocoapods-core/specification.rb
```diff
@@ -69,11 +69,6 @@ def initialize_copy(other)
     #
     attr_accessor :subspecs
 
-    # @return [Bool] If this specification is a test specification.
-    #
-    attr_accessor :test_specification
```
segiddins: removing this writer method is definitely a breaking change

lib/cocoapods-core/specification/json.rb
```diff
@@ -76,6 +76,17 @@ def self.from_hash(hash, parent = nil)
     def self.subspecs_from_hash(spec, subspecs)
       return [] if subspecs.nil?
       subspecs.map do |s_hash|
+        # We've already have the JSON here and there is no 'type' property.
+        if s_hash['type'].nil?
+          # Backwards compatibility with subspecs and test specs.
+          s_hash['type'] = if s_hash['test_type'].nil?
```
segiddins: could use `s_hash['type'] ||=` here

PR #462, opened 2018-09-13
Implement app specs

dostrander:


dostrander:
@segiddins @dnkoutso Ready for review

lib/cocoapods-core/specification/json.rb
```diff
@@ -22,17 +22,19 @@ def to_pretty_json(*a)
       #
       def to_hash
         hash = attributes_hash.dup
+        puts "parent: #{parent}"
```
dnkoutso: is this needed?

dostrander: Nope just forgot to remove

lib/cocoapods-core/specification.rb
```diff
@@ -37,13 +37,18 @@ class Specification
     # @param [Bool] test_specification
     #        Whether the specification is a test specification
     #
-    def initialize(parent = nil, name = nil, test_specification = false)
+    # @param [Bool] app_specification
+    #        Whether the specification is an app specification
+    #
+    def initialize(parent = nil, name = nil, test_specification: false, app_specification: false)
```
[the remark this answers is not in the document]
dostrander: 👍

lib/cocoapods-core/specification/json.rb
```diff
@@ -61,22 +62,31 @@ def self.from_json(json)
     #
     # @return [Specification] the specification
     #
-    def self.from_hash(hash, parent = nil)
-      spec = Spec.new(parent)
+    def self.from_hash(hash, parent = nil, test_specification = false, app_specification = false)
```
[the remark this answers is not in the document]
dostrander: 👍

dostrander:
Makes sense i'll put up the CocoaPods PR as well. I was thinking i'd start it from Core then in CocoaPods but this makes sense to see how it interacts with the CocoaPods.

lib/cocoapods-core/specification.rb
```diff
@@ -223,6 +233,18 @@ def subspec?
 
     # @!group Dependencies & Subspecs
 
+    # @return [Bool] If this specification is a libary (app or test) specification.
```
amorde: similar comment here, I think the terminology is a bit confusing but maybe its just me

lib/cocoapods-core/specification.rb
```diff
@@ -223,6 +233,18 @@ def subspec?
...
+    #
+    def library_specification?
+      !app_specification? && !test_specification?
+    end
+
+    # @return [Bool] If this specification is not a libary (app or test) specification.
+    #
+    def non_library_specification?
+      !library_specification?
```
amorde: is this needed? seems like just negating `library_specification` should be fine even from an API perspective

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -1486,6 +1486,28 @@ def test_spec(name = 'Tests', &block)
...
+      #
+      # @example
+      #
+      #   Pod::Spec.new do |spec|
+      #     spec.name = 'NSAttributedString+CCLFormat'
+      #
+      #     spec.app_spec do |app_spec|
+      #       app_spec.source_files = 'NSAttributedString+CCLFormat.m'
+      #       app_spec.dependency 'AFNetworking'
+      #     end
```
amorde: Ah, ok I see what this is doing. I still think a bit more info would be helpful here but I have a better idea of what's going on based on this

lib/cocoapods-core/specification.rb
dostrander: I'll update these comments to make them a little more clear, thanks!

lib/cocoapods-core/specification.rb
dostrander: `non_library_specification` is pulled from more so i figure it was just convenience. Don't feel to strongly about this though

amorde: 👍

segiddins:
Did we decide to not add a `Specification#type` method (or something similar)? Otherwise, this looks 👍

dostrander:
forgot about the `Specification#type` will add and update cocoapods

dostrander:
@segiddins have been playing around with it and it doesn't really clean anything up directly in cocoapods. Do you have any suggestions specifically?

segiddins:
There was stuff where we do like 

```
if spec.app_spec?
  target.file_accessors.select { |fa| fa.spec.app_spec? }
elsif
...
```

That could become `target.file_accessors.select { |fa| fa.spec.type == spec.type }`.

There's also places where its

```
library_specs = specs.select(&:library_spec?)
test_specs = specs.select(&:test_spec?)
app_specs = specs.select(&:app_spec?)
```

that could become `specs.group_by(&:type)`, and then you have a hash with each of the 3 arrays

lib/cocoapods-core/specification.rb
```diff
@@ -221,8 +231,45 @@ def subspec?
 
     public
 
+    # Spec types supported by the
+    SUPPORTED_SPEC_TYPES = [:root, :sub, :app, :test].freeze
```
segiddins: imho `root + sub` should be combined into `library`

lib/cocoapods-core/specification.rb
```diff
@@ -74,6 +79,11 @@ def initialize_copy(other)
     attr_accessor :test_specification
     alias_method :test_specification?, :test_specification
 
+    # @return [Bool] If this specification is an app specification.
+    #
+    attr_accessor :app_specification
```
dnkoutso: I was stupid when I did test specs and this should probably be `attr_reader`

lib/cocoapods-core/specification.rb
```diff
@@ -221,8 +231,38 @@ def subspec?
 
     public
 
+    # Spec types supported by the
```
dnkoutso: incomplete doc

lib/cocoapods-core/specification/dsl.rb
dnkoutso: Agreed, this documentation will be reflected on the website. I would expand a bit more on the functionality and value app specs provide.

For example, app specs can be used as sample consumers of the library as examples people can follow on how to actually integrate the library.

dnkoutso:
You do not to commit ```spec/fixtures/vcr_cassettes/MasterSource_fetch.yml
spec/fixtures/vcr_cassettes/MasterSource_nofetch.yml```

files I dont think

spec/specification_spec.rb
```diff
@@ -396,6 +467,15 @@ module Pod
         ]
       end
 
+      it 'returns the app spec dependencies' do
+        app_spec = @spec.app_spec { |s| s.dependency 'OCMock' }
+        app_spec .dependencies.sort.should == [
```
dnkoutso: nit: extra space

lib/cocoapods-core/specification.rb
dostrander: These actually need to be accessor because of the way we convert to and from json

lib/cocoapods-core/specification/json.rb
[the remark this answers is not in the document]
segiddins: what happened to this?

PR #468, opened 2018-11-13
Only compare segments once

segiddins:


dnkoutso:
Changelog? I also assume tests cover this

PR #469, opened 2018-11-15
Add CDN repo Source to allow retrieving specs from a web URL

igor-makarov:
This PR complements [#8280 in the main repo](https://github.com/CocoaPods/CocoaPods/pull/8280). 

See all the details there.


lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,199 @@
...
+    end
+
+    # @return [Pathname] The directory where the specs are stored.
+    #
+    def specs_dir
+      @specs_dir ||= repo + 'Specs'
+    end
+
+    def download_file(partial_url)
+      Pod::UI.message "CDN: #{name} Relative path to download: #{partial_url}"
```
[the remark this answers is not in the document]
igor-makarov: I've tried doing that, but the multi-threading made it completely unreadable... I ended up deleting the first `.message` and just made off with the other ones.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,199 @@
...
+      etag = File.read("#{path}.etag") if File.exist?("#{path}.etag")
+      Pod::UI.message "CDN: #{name} Relative path: #{partial_url}, has ETag? #{etag}" unless etag.nil?
+      response = Faraday.get(file_remote_url) do |req|
+        req.headers['If-None-Match'] = etag unless etag.nil?
+      end
+
+      case response.status
+      when 304
+        Pod::UI.message "CDN: #{name} Relative path not modified: #{partial_url}"
+        FileUtils.touch path
```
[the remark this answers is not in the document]
igor-makarov: It's an optimization: any file modified or touched after the source initializes, is considered fresh enough to not request it again from the CDN. I will comment on that.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    # Returns pod names for given array of specification paths.
+    #
+    # @param  [Array<String>] spec_paths
+    #         Array of file path names for specifications. Path strings should be relative to the source path.
+    #
+    # @return [Array<String>] the list of the name of Pods corresponding to specification paths.
+    #
+    def pods_for_specification_paths(spec_paths)
+      # spec_paths.map do |path|
```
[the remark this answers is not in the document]
igor-makarov: You're right, this got left in by accident... ☺️

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,199 @@
...
+
+      return @versions_by_name[name] unless @versions_by_name[name].nil?
+
+      versions_file_name = pod_path_partial(name).join('index.txt')
+      download_file(versions_file_name)
+
+      loaders = []
+      @versions_by_name[name] ||= local_file(versions_file_name).map do |v|
+        basename = v.chomp
+        unless local_file_exist?(pod_path_partial(name).join("#{basename}/#{name}.podspec.json"))
```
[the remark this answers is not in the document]
igor-makarov: Ideally, it should. But there are like 4 of those in the master repo, so I suggest we leave them out at the moment.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,199 @@
...
+      if File.exist?(path) && @startup_time < File.mtime(path)
+        Pod::UI.message "CDN: #{name} Relative path: #{partial_url} modified during this run! Returning local"
+        return partial_url
+      end
+
+      FileUtils.mkdir_p "#{repo}/#{directory_name}"
+
+      etag = File.read("#{path}.etag") if File.exist?("#{path}.etag")
+      Pod::UI.message "CDN: #{name} Relative path: #{partial_url}, has ETag? #{etag}" unless etag.nil?
+      response = Faraday.get(file_remote_url) do |req|
```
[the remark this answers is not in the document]
igor-makarov: I didn't notice it was using one! Will change...

igor-makarov:
I've made (hopefully) all the requested changes.

lib/cocoapods-core/cdn_source.rb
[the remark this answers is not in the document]
segiddins: There should be none in the master repo, but other repos may have them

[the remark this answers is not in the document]
igor-makarov: Yeah, but other repos don't really need CDN, right?

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+    #
+    # @return [Pathname] The path of the specification.
+    #
+    def specification_path(name, version)
+      raise ArgumentError, 'No name' unless name
+      raise ArgumentError, 'No version' unless version
+
+      relative_podspec = pod_path_partial(name).join("#{version}/#{name}.podspec.json")
+      download_file(relative_podspec)
+      pod_path(name).join("#{version}/#{name}.podspec.json")
```
amorde: `"#{version}/#{name}.podspec.json"` could be pulled out into a variable for reuse

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+    #
+    def search(query)
+      unless specs_dir
+        raise Informative, "Unable to find a source named: `#{name}`"
+      end
+      if query.is_a?(Dependency)
+        query = query.root_name
+      end
+
+      found = download_file(pod_path_partial(query).join('index.txt'))
```
amorde: could this reference a constant containing the name of the index file?

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+
+      loaders = []
+      @versions_by_name[name] ||= local_file(versions_file_name) do |file|
+        file.map do |v|
+          basename = v.chomp
+          # Optimization: ensure all the podspec files at least exist. The correct one will get refreshed
+          # in #specification_path regardless.
+          unless local_file_exist?(pod_path_partial(name).join("#{basename}/#{name}.podspec.json"))
+            loaders << Concurrent::Promise.execute(:executor => @executor) do
+              download_file(pod_path_partial(name).join("#{basename}/#{name}.podspec.json"))
```
amorde: `"#{basename}/#{name}.podspec.json"` could be pulled out into a variable for reuse

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+    end
+
+    def local_file(partial_url)
+      File.open("#{repo}/#{partial_url}") do |file|
+        yield file if block_given?
+      end
+    end
+
+    def local_file_exist?(partial_url)
+      File.exist?("#{repo}/#{partial_url}")
```
amorde: please use `join`

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+
+        etag_new = response.headers['etag'].first if response.headers.include?('etag')
+        Pod::UI.message "CDN: #{name} Relative path downloaded: #{partial_url}, save ETag: #{etag_new}"
+        File.open(etag_path, 'w') { |f| f.write(etag_new) } unless etag_new.nil?
+        partial_url
+      end
+    end
+
+    def local_file(partial_url)
+      File.open("#{repo}/#{partial_url}") do |file|
```
amorde: please use `join`

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+        yield file if block_given?
+      end
+    end
+
+    def local_file_exist?(partial_url)
+      File.exist?("#{repo}/#{partial_url}")
+    end
+
+    def pod_path_partial(pod_name)
```
amorde: nit: `relative_pod_path`

amorde: also can this be private?

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+      if found
+        set = set(query)
+        set if set.specification_name == query
+      end
+    end
+
+    def git?
+      false
+    end
+  end
```
amorde: I think `update` will need to be overridden since the default implementation calls out to `git`

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+      versions_file_name = pod_path_partial(name).join('index.txt')
+      download_file(versions_file_name)
+
+      loaders = []
+      @versions_by_name[name] ||= local_file(versions_file_name) do |file|
+        file.map do |v|
+          basename = v.chomp
+          # Optimization: ensure all the podspec files at least exist. The correct one will get refreshed
+          # in #specification_path regardless.
+          unless local_file_exist?(pod_path_partial(name).join("#{basename}/#{name}.podspec.json"))
```
amorde: Does this need to use the relative path? Ie. would `pod_path(name).join("#{basename}/#{name}.podspec.json").exist?` not work in this scenario?

lib/cocoapods-core/cdn_source.rb
amorde: `pod_path_partial(name)` is also duplicated a few times

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,210 @@
...
+    # @param  [String] name
+    #         the name of the Pod.
+    #
+    def versions(name)
+      return nil unless specs_dir
+      raise ArgumentError, 'No name' unless name
+
+      return @versions_by_name[name] unless @versions_by_name[name].nil?
+
+      versions_file_name = pod_path_partial(name).join('index.txt')
```
amorde: nit: `versions_file_path` since this is a path, not just a file name

lib/cocoapods-core/cdn_source.rb
amorde: actually now I'm wondering if `relative_pod_path` should be added to the main `Source` class

lib/cocoapods-core/cdn_source.rb
igor-makarov: I checked, and we don't actually get to CDNSource#update. The source manager says it's not a git repo, which is true. Luckily for us, there's no updating required for this repo type!

igor-makarov:
So I've picked all the nits, and made most of the implementation-related methods private.

I've begun adding unit tests. I'm using `webrick` for mocking the CDN server. I had to put its hookup in `spec_helper` to ensure no tests fail. I think it works better than VCR, because what I need is a read-only, static-hosted HTTP server that supports ETag out of the box.

igor-makarov:
I've added the tests for `CDNSource`. It's now 93% covered, according to what I can see.

spec/fixtures/mock_cdn_repo_remote/CocoaPods-version.yml
```diff
@@ -0,0 +1,7 @@
+---
```
segiddins: is it possible to use one of the existing fixture repos for this?

igor-makarov: I've briefly considered this. 
However, this is the mock of the *remote* end of the CDN source, and so cannot be in the `spec-repos` dir. Also, CDNSource wasn't designed to be backward-compatible with non-sharded repos.
Is there a problem with having a new fixture for a new source type?

igor-makarov: Also, I can reduce the number of files in it if necessary.

igor-makarov: Ok, I just tried adapting the existing fixtures to the new test suite. The sharding was not the issue but it didn't go well due to the podspecs not being JSON.

Currently, I have not added RB podspec support because it seems out of scope to me. The only truly huge spec repo is the main one, and it's JSON-only.

What is the main pain point of adding a new test fixture?

igor-makarov:
I've pruned the new `mock_cdn_repo_remote` fixture to have significantly less files. Hope that's better.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,270 @@
...
+    #
+    def specs_dir
+      @specs_dir ||= repo + 'Specs'
+    end
+
+    # @!group Querying the source
+    #-------------------------------------------------------------------------#
+
+    # @return [Array<String>] the list of the name of all the Pods.
+    #
```
dnkoutso: nit remove extra #

spec/fixtures/mock_cdn_repo_remote/all_pods.txt
```diff
@@ -0,0 +1,2 @@
+BeaconKit
+SDWebImage
```
amorde: will this file grow to be too large? not this spec specifically but on trunk. wondering if it should be split up

igor-makarov: [It's just 800KB now](
https://github.com/igor-makarov/Specs/blob/master/all_pods.txt). 

In fact, similar to Bundler's compact index, this saves time by coalescing it into jist one CDN call.

By the way, this only gets used in searching. If you specify `pod 'DfPodTest'` it will optimistically go for the specific pod index.

amorde: 👍

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,269 @@
...
+    # @param [String] repo The name of the repository
+    #
+    def initialize(repo)
+      # Optimization: we initialize startup_time when the source is first initialized
+      # and then test file modification dates against it. Any file that was touched
+      # after the source was initialized, is considered fresh enough.
+      @startup_time = Time.new
+
+      @executor = Concurrent::ThreadPoolExecutor.new(
+        :min_threads => 50,
```
segiddins: min_threads should probably be 1, from my experiences working on bundler, not every system is able to spin up 50 threads

lib/cocoapods-core/cdn_source.rb
```diff
@@ -0,0 +1,269 @@
...
+        unless repo.exist?
+          raise Informative, "Unable to find a source named: `#{name}`"
+        end
+
+        specs_dir.mkpath
+        download_file('CocoaPods-version.yml')
+
+        # Optimization: all the version index files in the local cache will in all probability be
+        # requested anyaway, better preload them in parallel.
+        glob_pattern = repo.join('*/**/*.txt')
```
segiddins: let's extract this into a separate method

igor-makarov:
I've made the changes requested by @segiddins.

PR #471, opened 2018-11-18
[Podfile] Add support for `Podfile.rb` file name

amorde:
For https://github.com/CocoaPods/CocoaPods/issues/8171

### DIFF
--- a/spec/podfile_spec.rb
+++ b/spec/podfile_spec.rb
@@ -415,6 +415,13 @@ module Pod
         Podfile.from_file(path)
       end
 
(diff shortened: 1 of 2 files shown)


spec/podfile_spec.rb
```diff
@@ -415,6 +415,13 @@ module Pod
         Podfile.from_file(path)
       end
 
+      it 'handles the `podfile` extension' do
```
amorde: whoops, copy-paste mistake

amorde: fixed

PR #473, opened 2018-11-28
Explain Handling of dash-separated Version Components

herzi:
* lib/cocoapods-core/podfile/dsl.rb: Extend the version explanations
  to cover the behavior for more complex version numbering schemes
  within semantic versioning.


herzi:
I wasn't sure about re-indenting the other blocks. If that needs to be done, just tell me. For now I tried to keep the delta as small as possible.

amorde:
Thanks @herzi! 👍

PR #474, opened 2018-11-30
Improve error messages  for podspec dependencies

Lutzifer:
Improves the error message ’Unsupported version requirements' if it is caused by specifying :git or :path in the podspec file.

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -680,11 +680,11 @@ module Pod
           version_requirements.each do |requirement|
             if requirement.is_a?(Hash)
               if !requirement[:path].nil?
-                raise Informative, 'Podspecs cannot specify the source of dependencies. The :path option is not supported.'\
-                                   ' :path can be used in the Podfile instead to override global dependencies.'
+                raise Informative, 'Podspecs cannot specify the source of dependencies. The `:path` option is not supported.'\
+                                   ' `:path` can be used in the Podfile instead to override global dependencies.'
               elsif !requirement[:git].nil?
-                raise Informative, 'Podspecs cannot specify the source of dependencies. The :git option is not supported.'\
-                                   ' :git can be used in the Podfile instead to override global dependencies.'
+                raise Informative, 'Podspecs cannot specify the source of dependencies. The `:git` option is not supported.'\
+                                   ' `:git` can be used in the Podfile instead to override global dependencies.'
               end
             end
           end
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -679,15 +679,15 @@ module Pod
         unless version_requirements.all? { |req| req.is_a?(String) }
           version_requirements.each do |requirement|
             if requirement.is_a?(Hash)
-              if requirement[:path] != nil
-                raise Informative, "Podspecs can only use remote pods as dependencies. :path is not supported"
-              elsif requirement[:git] != nil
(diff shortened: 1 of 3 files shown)


lib/cocoapods-core/specification/dsl.rb
```diff
@@ -677,7 +677,13 @@ def dependency(*args)
           end
         end
         unless version_requirements.all? { |req| req.is_a?(String) }
-          raise Informative, 'Unsupported version requirements'
+          unless version_requirements.is_a?(Hash)
+            version_requirements.each do |requirement|
+              raise Informative, "Podspecs can only use remote pods as dependencies. :path is not supported" if requirement[:path] != nil
```
dnkoutso: would prefer `if/elseif` here

dnkoutso:
Needs tests for each path and a changelog entry

orta:
Nice idea 👍

Lutzifer:
@dnkoutso @orta 
Thanks for the input. I think I solved the issues, could you please have another look?

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -677,8 +677,19 @@ def dependency(*args)
           end
         end
         unless version_requirements.all? { |req| req.is_a?(String) }
-          raise Informative, 'Unsupported version requirements'
+          version_requirements.each do |requirement|
+            if requirement.is_a?(Hash)
+              if !requirement[:path].nil?
+                raise Informative, 'Podspecs can only use remote pods as dependencies. :path is not supported'
+              elsif !requirement[:git].nil?
+                raise Informative, 'Podspecs can only use remote pods as dependencies. :git is not supported'
```
amorde: The error message isn't _technically_ correct - you can use `:git` dependencies, but they cannot be specified by the podspec itself (ex. if the pod file contains `pod 'MyPod', :git => '...'`, a podspec with the dependency `'MyPod'` will be fulfilled by that installation)

What about the following:

> Podspecs cannot specify the source of dependencies. The `:git` option is not supported

Lutzifer: After rereading this for a couple of times, I think In understand what you mean.

Will change the message.

Lutzifer: Is it the same with path?

amorde: yup `:path` behaves the same as far as I'm aware

amorde:
LGTM once lint warnings are addressed

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -677,8 +677,21 @@ def dependency(*args)
           end
         end
         unless version_requirements.all? { |req| req.is_a?(String) }
-          raise Informative, 'Unsupported version requirements'
+          version_requirements.each do |requirement|
+            if requirement.is_a?(Hash)
+              if !requirement[:path].nil?
+                raise Informative, 'Podspecs cannot specify the source of dependencies. The :path option is not supported.'\
```
dnkoutso: nit can you wrap :path with backticks \`\`? Same for :git below

dnkoutso:
ping @Lutzifer needs a rebase and a comment addressed

lib/cocoapods-core/specification/dsl.rb
Lutzifer: done

Lutzifer:
implemented the nit and did the rebase

amorde:
Thanks @Lutzifer!

PR #481, opened 2018-12-16
Source::Manager - return all aggregate sources if the dependency's spec repo doesn't exist

igor-makarov:
This PR resolves a bug I was having because I've switched to the new [CDN-source](https://github.com/CocoaPods/Core/pull/469) and had an error.

The error happened because while the Podfile had the new source specified, the Podfile.lock still had the old source and I have deleted the old source.  
I've gotten around it by manually hacking the lockfile, but clearly this isn't the right way.

I've modified the `aggregate_for_dependency` method in a way that preserves current behavior but in the case of not being able to find the specified source for the dependency, it will default to all sources. 

I've added a test for the case.

### DIFF
--- a/lib/cocoapods-core/source/manager.rb
+++ b/lib/cocoapods-core/source/manager.rb
@@ -33,15 +33,12 @@ module Pod
       #         will be used.
       #
       def aggregate_for_dependency(dependency)
-        if dependency.podspec_repo
-          source = source_with_url(dependency.podspec_repo)
-          raise StandardError, '[Bug] Failed to find known source with the URL ' \
-            "#{dependency.podspec_repo.inspect}" if source.nil?
+        return aggregate if dependency.podspec_repo.nil?
 
-          aggregate_with_repos([source.repo])
-        else
-          aggregate
-        end
+        source = source_with_url(dependency.podspec_repo)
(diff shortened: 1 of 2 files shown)


igor-makarov:
@segiddins Well, the PR is changing _some_ behavior.

What I meant to say is that existing *expected* behavior doesn't change.  
What happened in my case was this:
1. I've added the new CDNSource.
2. Realized that because of the Lockfile, new CDNSource wasn't being used.
3. Deleted the master repo from disk.
4. Got an error saying 'Couldn't find repo `https://github.com/CocoaPods/Specs.git`'.

My expectation was that the installer will try to find all the locked dependencies in the new repo.  
The PR does just that, and I think this is a reasonable expectation in the case of switching spec repos. It doesn't in fact change any version locks, just the spec repo URL in the Lockfile.

The alternative to this is hacking the Lockfile by hand and I don't think this is something we want all users to do when CDNSource goes mainstream.

dnkoutso:
@igor-makarov needs a rebase and we can merge this.

igor-makarov:
@dnkoutso Rebased! 💪

igor-makarov:
🎉

PR #511, opened 2019-02-03
Change CDNSource to be optimistic

igor-makarov:
I've been testing CDNSource myself for a while now and I think that I can make it a bit more optimistic.

When running `pod install`, it will assume that all existing files do not need updating.

To update the existing files, a run of `pod repo update` will be needed, but `pod update` runs it by default.

### DIFF
--- a/lib/cocoapods-core/cdn_source.rb
+++ b/lib/cocoapods-core/cdn_source.rb
@@ -211,6 +211,10 @@ module Pod
     end
 
     def git?
+      # Long story here. This property is actually used solely by Source::Manager to determine
+      # which sources are updatable. Ideally, this would require a name change but @segiddins
+      # has pointed out that it is public and could break plugins.
+      # In any case, CDN-backed repos can be updated and therefore the value ought to be true.
       true
     end
 
--- a/lib/cocoapods-core/cdn_source.rb
+++ b/lib/cocoapods-core/cdn_source.rb
@@ -9,6 +9,7 @@ module Pod
     # @param [String] repo The name of the repository
     #
     def initialize(repo)
+      @check_existing_files_for_update = false
       # Optimization: we initialize startup_time when the source is first initialized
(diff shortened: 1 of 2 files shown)


lib/cocoapods-core/cdn_source.rb
```diff
@@ -195,20 +192,26 @@ def search_by_name(query, full_text_search = false)
...
+        preheat_existing_files
+      ensure
+        @check_existing_files_for_update = false
+      end
       []
     end
 
     def git?
-      false
+      true
```
dnkoutso: is this change correct?

igor-makarov: Well, yes and no. 🙃

Yes, because the CDN source this is modeled on is a mirror of a git repo.

No, because it's not technically a git repo.

Personally, according to all the uses, I think the property should be called `can_be_updated?` on the base class, but this will require another PR in main.

segiddins: Probably should be broken out into 2 different atrtibutes, `git?` and `updatable?`, and the default implementation for `updatable?` can return `git?`

igor-makarov: Hehe... However from what I've seen there are no other uses of `git?`, so YAGNI 🤷🏻‍♂️

segiddins: It is a public method though, and I'm worried about it being a breaking change (potentially) for other source plugins

dnkoutso: >  Yes, because the CDN source this is modeled on is a mirror of a git repo. No, because it's not technically a git repo.

Worthwhile for a code comment actually.

igor-makarov:
I've added the comment on `git?` property. 😂

igor-makarov:
💪

PR #520, opened 2019-03-21
Add `test_spec. app_host_name` to DSL

jkap:
This supports a forthcoming PR in cocoapods/cocoapods to specify an
existing app spec to use as a test spec app host instead of generating
one on the fly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@
 
 ##### Enhancements
 
-* Add `test_spec.app_spec_name` to DSL  
+* Allow test specs to specify an `app_host_name` to point to an app spec whose
+  application is used as the test's app host.  
   [jkap](https://github.com/jkap)
+  [Samuel Giddins](https://github.com/segiddins)
   [#520](https://github.com/CocoaPods/Core/pull/520)
 
 ##### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Add `test_spec.app_spec_name` to DSL  
+  [jkap](https://github.com/jkap)
+  [#520](https://github.com/CocoaPods/Core/pull/520)
 
 ##### Bug Fixes
 
--- a/lib/cocoapods-core/specification/consumer.rb
+++ b/lib/cocoapods-core/specification/consumer.rb
@@ -144,6 +144,10 @@ module Pod
       spec_attr_accessor :requires_app_host
       alias_method :requires_app_host?, :requires_app_host
 
+      # @return [String?] Name of the app host this spec requires
+      #
+      spec_attr_accessor :app_host_name
+
       # @return [Symbol] the test type supported by this specification.
       #
       spec_attr_accessor :test_type
--- a/lib/cocoapods-core/specification/consumer.rb
+++ b/lib/cocoapods-core/specification/consumer.rb
@@ -144,7 +144,7 @@ module Pod
       spec_attr_accessor :requires_app_host
       alias_method :requires_app_host?, :requires_app_host
 
-      # @return [String?] Name of the app host this spec requires
+      # @return [String] Name of the app host this spec requires
       #
       spec_attr_accessor :app_host_name
 
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1511,10 +1511,24 @@ module Pod
       # @!method app_host_name=(name)
       #
       #   The app specification to use as an app host, if necessary.
+      #   Requires `test_spec.requires_app_host` to be `true`.
       #
-      #   @example
+      # @example
       #
-      #     test_spec.app_host_name = 'Foo'
+      #   Pod::Spec.new do |spec|
+      #     spec.name = 'NSAttributedString+CCLFormat'
+      #
+      #     spec.test_spec do |test_spec|
+      #       test_spec.source_files = 'NSAttributedString+CCLFormatTests.m'
+      #       test_spec.requires_app_host = true
+      #       test_spec.app_host_name = 'NSAttributedString+CCLFormat/App'
+      #     end
+      #
+      #     spec.app_spec 'App' do |app_spec|
+      #       app_spec.source_files = 'NSAttributedString+CCLFormat.m'
+      #       app_spec.dependency 'AFNetworking'
+      #     end
+      #   end
       #
       #   @param [String] name
       #          The app specification to use as an app host, if necessary.
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1508,6 +1508,21 @@ module Pod
                 :default_value => false,
                 :spec_types => [:test]
 
+      # @!method app_host_name=(name)
+      #
+      #   The app specification to use as an app host, if necessary.
+      #
+      #   @example
+      #
+      #     test_spec.app_host_name = 'Foo'
+      #
+      #   @param [String] name
+      #          The app specification to use as an app host, if necessary.
+      #
+      attribute :app_host_name,
+                :types => [String],
+                :spec_types => [:test]
+
       SCHEME_KEYS = [:launch_arguments, :environment_variables].freeze
 
       # @!method scheme=(flag)
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1512,7 +1512,9 @@ module Pod
       #
       #   The app specification to use as an app host, if necessary.
       #
-      #   **Note** If the app host is in another pod, you must depend on that pod using `test_spec.dependency 'PodName'`
+      # @note
+      #
+      #    You must depend on that app spec using `test_spec.dependency 'PodName'`.
       #
       # @example
       #
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1511,8 +1511,8 @@ module Pod
       # @!method app_host_name=(name)
       #
       #   The app specification to use as an app host, if necessary.
-      #   
-      #   **Note** Requires `test_spec.requires_app_host` to be `true`.
+      #
+      #   **Note** If the app host is in another pod, you must depend on that pod using `test_spec.dependency 'PodName'`
       #
       # @example
       #
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1511,7 +1511,8 @@ module Pod
       # @!method app_host_name=(name)
       #
       #   The app specification to use as an app host, if necessary.
-      #   Requires `test_spec.requires_app_host` to be `true`.
+      #   
+      #   **Note** Requires `test_spec.requires_app_host` to be `true`.
       #
       # @example
       #
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1525,6 +1525,7 @@ module Pod
       #       test_spec.source_files = 'NSAttributedString+CCLFormatTests.m'
       #       test_spec.requires_app_host = true
       #       test_spec.app_host_name = 'NSAttributedString+CCLFormat/App'
+      #       test_spec.dependency 'NSAttributedString+CCLFormat/App'
       #     end
       #
       #     spec.app_spec 'App' do |app_spec|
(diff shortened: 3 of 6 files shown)


lib/cocoapods-core/specification/consumer.rb
```diff
@@ -144,6 +144,10 @@ def user_target_xcconfig
       spec_attr_accessor :requires_app_host
       alias_method :requires_app_host?, :requires_app_host
 
+      # @return [String?] Name of the app host this spec requires
```
dnkoutso: nit remove `?`

jkap: is there a preferred way to specify that this is an optional?

dnkoutso: I would expand on documentation. This flag is only applied if `requires_app_host` is set to `true`.

Also `app_host_name` does not convey a message that this can point to an app spec name. Perhaps include that in documentation?

The docs in here are actually mapped into the official docs.

dnkoutso: Oops I am super duper wrong this is `consumer.rb`

dnkoutso: Please disregard most of this as I see below it does specify app specs.

spec/specification/consumer_spec.rb
```diff
@@ -289,6 +289,14 @@ module Pod
         test_consumer.requires_app_host?.should.be.true
       end
 
+      it 'allows to specify a specific app host to use' do
```
dnkoutso: what happens if we specify an app host name that is not declared within the podspec?

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -1508,6 +1508,21 @@ def subspec(name, &block)
                 :default_value => false,
                 :spec_types => [:test]
 
+      # @!method app_host_name=(name)
```
dnkoutso: Should we specify that `requires_app_host` is required for this to work? is it true?

spec/specification/consumer_spec.rb
dnkoutso: if we require `requires_app_host` to be true in order for this flag to function we probably want a test for it.

lib/cocoapods-core/specification/consumer.rb
dnkoutso: remove `?`

jkap: my original question about specifying this is an option still stands

spec/specification/consumer_spec.rb
jkap: at the moment that checking is done in main cocoapods during integration, wasn't sure the best place to do that validation in core.

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -1508,6 +1508,21 @@ def subspec(name, &block)
...
+      #
+      #   The app specification to use as an app host, if necessary.
+      #
+      #   @example
+      #
+      #     test_spec.app_host_name = 'Foo'
```
dnkoutso: lets provide a full example as the test includes.

```
      # @example
      #
      #   Pod::Spec.new do |spec|
      #     spec.name = 'AFNetworking'
      #
      #     spec.app_spec 'App' do |app_spec|
      #       app_spec.source_files = 'NSAttributedString+CCLFormat.m'
      #     end
      #     
      #     spec.test_spec do |test_spec|
      #       test_spec.app_host_name = 'AFNetworking/App'
      #     end
      #
      #   end
```

spec/specification/consumer_spec.rb
dnkoutso: you can probably perform it at this level check `linter.rb` in Core

lib/cocoapods-core/specification/linter.rb
```diff
@@ -394,6 +394,13 @@ def _validate_test_type(t)
         end
       end
 
+      def _validate_app_host_name(n)
+        unless @consumer.requires_app_host?
```
dnkoutso: nit `consumer`

lib/cocoapods-core/specification/consumer.rb
dnkoutso: hm not sure what consumer does in the case its empty but I haven't seen this convention for documentation anywhere else.

spec/specification/consumer_spec.rb
dnkoutso: I believe this is still missing a test case that checks that the value of `app_host_name` actually matches an app spec. If it doesn't we should probably error out.

lib/cocoapods-core/specification/dsl.rb
segiddins: we should, since the validator checks for it

spec/specification/consumer_spec.rb
segiddins: since the app spec can be from another pod, there's no immediate way to validate that

lib/cocoapods-core/specification/linter.rb
```diff
@@ -394,6 +394,13 @@ def _validate_test_type(t)
         end
       end
 
+      def _validate_app_host_name(_n)
```
segiddins: should we also validate there's a dependency on a pod with that name, to make sure it's also being being?

lib/cocoapods-core/specification/dsl.rb
segiddins: alternatively, should setting `app_host_name` add a dependency to that spec and also enable `requires_app_host`?

spec/specification/consumer_spec.rb
dnkoutso: wow it can be from another pod?! I did not know this was part of the design. This seems pretty bad imo, how will the validator (lint) deal with this? Do we implicitly add a `s.dependency`?

jkap: I don't like doing things implicitly, but I am adding a check to verify that we depend on a pod with the right name. If the app host doesn't exist, that'll come up during integration, but I don't know of a way to error out of that earlier.

lib/cocoapods-core/specification/linter.rb
```diff
@@ -394,6 +394,21 @@ def _validate_test_type(t)
...
+        unless consumer.requires_app_host?
+          results.add_error('app_host_name', '`requires_app_host` must be set to ' \
+            '`true` when `app_host_name` is specified.')
+        end
+
+        parent_spec_name = (consumer.spec.parent && consumer.spec.parent.name) || consumer.name
+        podspec_name, appspec_name = n.split('/')
+
+        unless podspec_name == parent_spec_name || consumer.dependencies.map(&:name).include?(appspec_name)
+          results.add_error('app_host_name', "Test spec must depend on `#{podspec_name}` " \
```
dnkoutso: message can include `Test spec `#{consumer.spec.name}` ...` so the user knows explicitly which test spec has the problem.

Alternatively the message can be "The app host name specified by `#{consumer.spec.name}` could not be found. Either explicitly declare it within `#{podspec_name}` or add a dependency to the podspec that specifies it."

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -1508,6 +1508,35 @@ def subspec(name, &block)
                 :default_value => false,
                 :spec_types => [:test]
 
+      # @!method app_host_name=(name)
+      #
+      #   The app specification to use as an app host, if necessary.
+      #   Requires `test_spec.requires_app_host` to be `true`.
```
dnkoutso:
```ruby
      #   The app specification to use as an app host, if necessary.

      #   **Note** Requires `test_spec.requires_app_host` to `true`.
```

basically add the `**Note**` which is a pattern across this file.

lib/cocoapods-core/specification/dsl.rb
segiddins: I think it's a bad user experience to require 2 lines to be written by rote whenever you have another one, when the user's intent is clear

jkap: i think having `requires_app_host` be auto-set makes sense but I would really rather not auto-require dependencies since that reduces visibility into what's actually happening

jkap:
See cocoapods/cocoapods#8654

lib/cocoapods-core/specification/linter.rb
```diff
@@ -394,6 +394,22 @@ def _validate_test_type(t)
...
+      def _validate_app_host_name(n)
+        unless consumer.requires_app_host?
+          results.add_error('app_host_name', '`requires_app_host` must be set to ' \
+            '`true` when `app_host_name` is specified.')
+        end
+
+        parent_spec_name = (consumer.spec.parent && consumer.spec.parent.name) || consumer.name
+        podspec_name, appspec_name = n.split('/')
+
+        unless podspec_name == parent_spec_name || consumer.dependencies.map(&:name).include?(appspec_name)
```
segiddins: This doesn't work when depending upon an app spec in another pod, since the dependency will contain the fully-qualified name rather than just the name of the appspec

spec/specification/linter_spec.rb
```diff
@@ -305,6 +305,38 @@ def result_should_include(*values)
...
+      end
+
+      it 'fails a test spec requiring an app host from a pod that isn\'t required' do
+        @spec.test_specification = true
+        @spec.requires_app_host = true
+        @spec.app_host_name = 'Foo/App'
+        result_should_include('app_host_name', 'Foo')
+      end
+
+      it 'passes a test spec requiring an app host from a pod that is listed as a dependency' do
```
segiddins: this test doesnt actually assert that the linter passes, rather it asserts that there is an error

spec/specification/consumer_spec.rb
dostrander: IMO we shouldn't add a dependency implicitly however I don't mind the fully qualified name (i.e. `$PodspecName/$AppSpecName`. One thing that i can see as a gotcha (super edge case) is if a podspec or the app specs has a `/` in the name? I'm not sure if the cocoapods validator doesn't allow that or not however if it does it could make it so there is 2 app specs with the same fully qualified name.

dnkoutso: > I'm not sure if the cocoapods validator doesn't allow that or not however if it does it could make it so there is 2 app specs with the same fully qualified name.

I do not believe it is allowed for root specs.

dnkoutso: See https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/specification/linter.rb#L203-L206

lib/cocoapods-core/specification/linter.rb
```diff
@@ -394,6 +394,18 @@ def _validate_test_type(t)
...
       end
 
+      def _validate_app_host_name(n)
+        unless consumer.requires_app_host?
+          results.add_error('app_host_name', '`requires_app_host` must be set to ' \
+            '`true` when `app_host_name` is specified.')
+        end
+
+        unless consumer.dependencies.map(&:name).include?(n)
+          results.add_error('app_host_name', "The app host name (#{n}) specified by #{consumer.spec.name} could " \
```
dnkoutso: nit if you care wrap `#{consumer.spec.name}` with backticks

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Add `test_spec.app_spec_name` to DSL
```
dnkoutso: nit this should say `app_host_name`

lib/cocoapods-core/specification/linter.rb
dnkoutso: nit still applies.

CHANGELOG.md
```diff
@@ -4,7 +4,11 @@
 
 ##### Enhancements
 
-* None.  
+* Allow test specs to speficy an `app_host_name` to point to an app spec whose
```
dnkoutso: s/speficy/specify

segiddins: Fixed

lib/cocoapods-core/specification/dsl.rb
```diff
@@ -1508,6 +1508,38 @@ def subspec(name, &block)
...
+      #
+      # @example
+      #
+      #   Pod::Spec.new do |spec|
+      #     spec.name = 'NSAttributedString+CCLFormat'
+      #
+      #     spec.test_spec do |test_spec|
+      #       test_spec.source_files = 'NSAttributedString+CCLFormatTests.m'
+      #       test_spec.requires_app_host = true
+      #       test_spec.app_host_name = 'NSAttributedString+CCLFormat/App'
```
dnkoutso: `test_spec.dependency` is required here. The example should denote that.

PR #524, opened 2019-04-03
delete search index json file when parsing failed

tripleCC:
for issue https://github.com/CocoaPods/CocoaPods/issues/5172

I think it is more friendly to delete `search_index.json` file  automatically when we fail to parse the file.

for now,  I have to delete `search_index.json` file manually every time when I encounter the problem.

It's inconvenient especially when I forget where is the `search_index.json` file located.


segiddins:
Could you please add a test for this change? Thanks!

dnkoutso:
If you want it for 1.7 then please target the `1-7-stable` branch.

PR #536, opened 2019-05-09
Produce an error when a Podspec contains an invalid type for the `source` attribute

amorde:
Closes CocoaPods/CocoaPods#5420

### DIFF
--- a/spec/specification/linter_spec.rb
+++ b/spec/specification/linter_spec.rb
@@ -176,9 +176,14 @@ module Pod
         result_should_include('license', 'unrecognized `name` key')
       end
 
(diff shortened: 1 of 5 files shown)


spec/specification/linter_spec.rb
```diff
@@ -176,9 +176,14 @@ def result_should_include(*values)
         result_should_include('license', 'unrecognized `name` key')
       end
 
-      xit 'checks the source for unknown keys' do
```
amorde: this test has been disabled for a long time, decided to update it while I was here

segiddins:
Nice! 👍🏻

PR #551, opened 2019-06-17
Improve threading in CDNSource

igor-makarov:
This PR addresses several issues in `CDNSource`:
1. Thread exhaustion on low-performance machines (such as CI VMs). The default number of threads is now set to 50 with an environment variable allowing to set it differently (`MAX_CDN_NETWORK_THREADS`).
2. Mangled output when concurrent code was failing. Now it will provide a more structured and readable output. Also using `Future`s as apparently this is what the `ruby-concurrent` docs suggest.
3. Non-HTTP errors were raised as exceptions. Now these will be wrapped with data about the specific download.
4. Added retry for network-related errors (`MAX_NUMBER_OF_RETRIES = 5`).

### DIFF
--- a/lib/cocoapods-core/cdn_source.rb
+++ b/lib/cocoapods-core/cdn_source.rb
@@ -6,6 +6,8 @@ module Pod
   # Subclass of Pod::Source to provide support for CDN-based Specs repositories
   #
   class CDNSource < Source
+    MAX_NUMBER_OF_RETRIES = 5
+
     # @param [String] repo The name of the repository
     #
     def initialize(repo)
@@ -309,11 +311,7 @@ module Pod
       path = repo + partial_url
       etag_path = path.sub_ext(path.extname + '.etag')
 
-      begin
-        response = etag.nil? ? REST.get(file_remote_url) : REST.get(file_remote_url, 'If-None-Match' => etag)
-      rescue => e
-        raise Informative, "CDN: #{name} Relative path couldn't be downloaded: #{partial_url}, error: #{e}"
-      end
+      response = download_retrying_connection_errors(partial_url, file_remote_url, etag)
 
       case response.status_code
       when 301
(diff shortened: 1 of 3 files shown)


dnkoutso:
Do we want this for 1.7.3 or 1.8.0?

dnkoutso:
If 1.7 then point to `1-7-stable` branch and include `CHANGELOG` entry.

igor-makarov:
You're right, 1.7 makes more sense.  
Do I need to make a new PR for that?

dnkoutso:
No. `git rebase` your branch against `1-7-stable`, push it, then come here and update the target branch at the top to point to `1-7-stable` instead of `master`

igor-makarov:
Aaah, found the pointing picker. Coming up!

dnkoutso:
How hard/easy is to write tests for this change?

igor-makarov:
I rebased/repointed. Will test for the error messages but I don't know how to test the thread count.

igor-makarov:
I ended up adding network error retries to the same PR cause it was small extra effort.

spec/cdn_source_spec.rb
```diff
@@ -71,8 +71,48 @@ def all_local_files
       it 'raises if unexpected HTTP error' do
         REST.expects(:get).returns(REST::Response.new(500))
         should.raise Informative do
-          @source.versions('Unknown_Pod')
-        end.message.should.match /CDN: .* Relative path couldn\'t be downloaded: .* Response: 500/
+          @source.specification('BeaconKit', '1.0.0')
+        end.message.
+          should.== "CDN: #{@source.name} Relative path couldn\'t be downloaded: Specs/2/0/9/BeaconKit/1.0.0/BeaconKit.podspec.json Response: 500"
```
amorde: nit: remove trailing `.` after `should`, a few other instances below as well

lib/cocoapods-core/cdn_source.rb
```diff
@@ -16,8 +18,8 @@ def initialize(repo)
       @startup_time = Time.new
 
       @executor = Concurrent::ThreadPoolExecutor.new(
-        :min_threads => 1,
-        :max_threads => 200,
+        :min_threads => 5,
+        :max_threads => (ENV['MAX_CDN_NETWORK_THREADS'] || 50).to_i,
```
amorde: could `50` be put into a constant similar to `MAX_NUMBER_OF_RETRIES`?

spec/cdn_source_spec.rb
igor-makarov: Funnily enough, the syntax for matching raised exceptions doesn't allow that. Very strange but you're welcome to try it yourself 😂

I changed the one place where it isn't a raised exception.

lib/cocoapods-core/cdn_source.rb
igor-makarov: Done.

igor-makarov:
Addressed @amorde's feedback.

dnkoutso:
For https://github.com/CocoaPods/CocoaPods/issues/8926

PR #555, opened 2019-07-02
Gracefully handle invalid version strings during linting

amorde:
Closes CocoaPods/CocoaPods#8785

Output of lint for the example on the above issue will now look like this (a bit ugly, but nicer than a crash)

```
 -> SharkORM ({VER})
    - ERROR | attributes: Unable to parse attribute `version` due to error: Malformed version number string {VER}
    - ERROR | version: Unable to validate due to exception: Malformed version number string {VER}
    - ERROR | source: Unable to validate due to exception: Malformed version number string {VER}
    - ERROR | attributes: Unable to validate `version` (Malformed version number string {VER}).
    - ERROR | [iOS] unknown: Encountered an unknown error (The `SharkORM` pod failed to validate due to 4 errors:
    - ERROR | attributes: Unable to parse attribute `version` due to error: Malformed version number string {VER}
    - ERROR | version: Unable to validate due to exception: Malformed version number string {VER}
    - ERROR | source: Unable to validate due to exception: Malformed version number string {VER}
    - ERROR | attributes: Unable to validate `version` (Malformed version number string {VER}).

) during validation.
```

### DIFF
--- a/lib/cocoapods-core/specification.rb
+++ b/lib/cocoapods-core/specification.rb
@@ -125,8 +125,9 @@ module Pod
     #         clients.
     #
     def to_s
-      if name && !version.version.empty?
-        "#{name} (#{version})"
+      specified_version = attributes_hash['version'] || ''
+      if name && !specified_version.empty?
+        "#{name} (#{specified_version})"
       elsif name
         name
       else
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@
   [Eric Amorde](https://github.com/amorde)
   [CocoaPods/CocoaPods#5420](https://github.com/CocoaPods/CocoaPods/issues/5420)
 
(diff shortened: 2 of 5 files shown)


lib/cocoapods-core/specification.rb
```diff
@@ -125,8 +125,9 @@ def hash
     #         clients.
     #
     def to_s
-      if name && !version.version.empty?
-        "#{name} (#{version})"
+      specified_version = attributes_hash['version'] || ''
+      if name && !specified_version.empty?
+        "#{name} (#{specified_version})"
```
amorde: needed because the validator calls `to_s` on specs, and this would crash on any invalid version string

dnkoutso:
Any tests possible to be added for this?

amorde:
Yeah I think tests should be simple, good call

PR #559, opened 2019-07-10
Add more logging to CDN source

cltnschlosser:
Extra logging for: https://github.com/CocoaPods/CocoaPods/issues/8987

### DIFF
--- a/spec/cdn_source_spec.rb
+++ b/spec/cdn_source_spec.rb
@@ -112,7 +112,7 @@ module Pod
         should.raise Informative do
           @source.specification('BeaconKit', '1.0.0')
         end.message.
-          should.== "CDN: #{@source.name} Relative path couldn\'t be downloaded: Specs/2/0/9/BeaconKit/1.0.0/BeaconKit.podspec.json Response: 500"
+          should.== "CDN: #{@source.name} Relative path couldn\'t be downloaded: #{@source.url}Specs/2/0/9/BeaconKit/1.0.0/BeaconKit.podspec.json Response: 500"
       end
 
       it 'raises if unexpected non-HTTP error' do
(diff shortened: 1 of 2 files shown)


dnkoutso:
Want to update the commits in this PR it seems it includes a merge commit to master

dnkoutso:
Test 'raises if unexpected HTTP error' failed.

Use `bundle exec bacon <filename>_spec.rb` to run tests locally.

cltnschlosser:
Yeah, I clicked the update button that github suggested :/

Okay, I'll run the tests locally

cltnschlosser:
Should be good now!

spec/cdn_source_spec.rb
```diff
@@ -112,7 +112,7 @@ def cleanup
         should.raise Informative do
           @source.specification('BeaconKit', '1.0.0')
         end.message.
-          should.== "CDN: #{@source.name} Relative path couldn\'t be downloaded: Specs/2/0/9/BeaconKit/1.0.0/BeaconKit.podspec.json Response: 500"
+          should.== "CDN: #{@source.name} Relative path couldn\'t be downloaded: #{@source.url}Specs/2/0/9/BeaconKit/1.0.0/BeaconKit.podspec.json Response: 500"
```
dnkoutso: its generally advisable not to use the actual code for assertions as you might end up with a false positive. Recommend putting the actual string here.

dnkoutso: Also I see that it does it in the beginning so maybe its fine.

cltnschlosser: Makes sense. I was just following the example of `#{@source.name}`.

PR #560, opened 2019-07-11
Improve CDN error reporting

igor-makarov:
#559 was a step in the right direction, however there are a few more additions to it that are needed:
* The actual wording of the messages didn't reflect the data presented.
* There was one other case that wasn't covered.
* Redirect message wasn't consistent with other messages or being tested.


igor-makarov:
Related to [CocoaPods#8987](https://github.com/CocoaPods/CocoaPods/issues/8987).

PR #561, opened 2019-07-13
Optimize repo update

igor-makarov:
Intended to go after [Specs#14418](https://github.com/CocoaPods/Specs/pull/14418) is merged.

This PR will reduce the amount of files requested when executing `pod repo update` on a CDN repo.

The files that actually need to be updated are the index files. The podspecs themselves aren't updateable except in these cases:
#### Deletion
When a podspec is deleted, it will disappear from the index file. I've patched a bug where `CDNSource` wasn't treating this correctly.
#### Deprecation
When a podspec is deprecated, it isn't reflected in the index. Since deprecations are pretty rare (5k podspecs), I've made the Netlify generator make a list of all the deprecated podspecs. This list is currently around 300KB and is expected to update very infrequently, so traffic would be very minimal.  
Upon updating, `CDNSource` will look if the user's repo dir has any of the deprecated files, and will add them to the update list. Since there are very few of those, this will not impact performance significantly.

In practice, the number of HTTP requests went down from thousands to around 50-100. While I expect this would increase traffic somewhat, the performance improvement for users, especially on non-ideal connections, is very good.


igor-makarov:
This was inspired by the discussion in [CocoaPods#8987](https://github.com/CocoaPods/CocoaPods/issues/8987) and @cltnschlosser's ideas 💪

cltnschlosser:
Seems like a low risk improvement.

Do you need special access to netlify for testing something like this? Or is there a workflow for using a local CDN that can use any Specs fork/branch? If it's the latter could you point to some documentation, or give me a quick overview. I wouldn't mind playing around with different indexes to see if I can measure improvements and tradeoffs to various approaches.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -56,8 +56,9 @@ def refresh_metadata
     end
 
     def preheat_existing_files
-      all_existing_files = [repo.join('**/*.yml'), repo.join('**/*.txt'), repo.join('**/*.json')].map(&Pathname.method(:glob)).flatten
-      loaders = all_existing_files.map { |f| f.relative_path_from(repo).to_s }.map do |file|
+      files_to_update = files_definitely_to_update + deprecated_local_podspecs
+      debug "CDN: #{name} Going to update #{files_to_update.count} files"
```
cltnschlosser: This is check for updates, right? Only actually doing the update if the etags don't match?

lib/cocoapods-core/cdn_source.rb
```diff
@@ -68,6 +69,21 @@ def preheat_existing_files
...
 
+    def files_definitely_to_update
+      [repo.join('**/*.yml'), repo.join('**/*.txt')].
+        map(&Pathname.method(:glob)).
+        flatten.
+        map { |f| f.relative_path_from(repo).to_s }
+    end
+
+    def deprecated_local_podspecs
+      download_file('deprecated_podspecs.txt')
```
cltnschlosser: Is there anyway to short circuit here if `deprecated_podspecs.txt` wasn't changed? Etag matched?

igor-makarov:
Netlify has generous free plans, so you can play with it yourself. Here's my personal build of the aforementioned Specs fork: 
https://app.netlify.com/sites/cocoapods-specs/deploys


The change in Specs builder on Netlify is functionally additive but the implementation has been rewritten so I'd like as much eyes on it:
@dnkoutso @segiddins @orta @amorde 

I've basically realized the the complexity of the builder has increased and I felt like leveraging existing CocoaPods capabilities. So the builder `require 'cocoapods'` now. There's a `Gemfile` and everything, which Netlify supports and caches.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -144,6 +160,10 @@ def versions(name)
     def specification_path(name, version)
       raise ArgumentError, 'No name' unless name
       raise ArgumentError, 'No version' unless version
+      unless versions(name).include?(Version.new(version))
```
cltnschlosser: This is the added code safe guarding against deleted podspecs?

lib/cocoapods-core/cdn_source.rb
igor-makarov: ETag is a server-side feature. You send an `if-none-match`, server either sends you a new file, or replies with `HTTP 304 Not Modified`.

lib/cocoapods-core/cdn_source.rb
igor-makarov: All downloads in update mode include the `if-none-match` header.

lib/cocoapods-core/cdn_source.rb
igor-makarov: Yes.

cltnschlosser:
> Netlify has generous free plans, so you can play with it yourself. Here's my personal build of the aforementioned Specs fork:
> https://app.netlify.com/sites/cocoapods-specs/deploys

So using `https://cocoapods-specs.netlify.com` as a CDN in Podfile, should be able to test backwards compatibility of your changes by using latest 1.7 release?

igor-makarov:
> > Netlify has generous free plans, so you can play with it yourself. Here's my personal build of the aforementioned Specs fork:
> > https://app.netlify.com/sites/cocoapods-specs/deploys
> 
> So using `https://cocoapods-specs.netlify.com` as a CDN in Podfile, should be able to test backwards compatibility of your changes by using latest 1.7 release?

Yeah, just don't forget the trailing slash, as the fix for it ins't in 1.7 😂

lib/cocoapods-core/cdn_source.rb
cltnschlosser: But do we always check all deprecated local podspecs? If `deprecated_podspecs.txt` `304`s, couldn't we just return `[]` here? Since deprecations are rare, this could be a big win.

igor-makarov: I don't quite follow what you mean. This method intersects the user's repo directory with the entire list of deprecations.

cltnschlosser: So if I understand what this does, it downloads `deprecated_podspecs.txt` if needed.
Then finds all locally tracked deprecated podspecs and calls `download_file` for them. Updating if needed (not 304), but a request is still made.

My thinking was if `deprecated_podspecs.txt` 304ed, then there is no reason to check all the locally known deprecated_podspecs for changes. In which case we can lower the network calls made even more.

cltnschlosser: Also I don't know how much `Core` knows about parsing podspecs, but taking a few lines from your Specs script update:
```
podspec = Pod::Specification.from_file(podspec_path)
podspec.deprecated?
```
We only need to make a request for locally tracked deprecated podspecs where the above is `false`.

cltnschlosser:
@igor-makarov For what it's worth, I had no backwards compatibility problems using your CDN with my work project using cocoapods 1.7.3.

lib/cocoapods-core/cdn_source.rb
igor-makarov: Theoretically, what you're saying sounds correct.  
However, I chose not to do this at this time since the expected value of these unnecessary updates is quite tiny. Deprecated podspecs are less than 2% of the total, and because they are deprecated, the chances of them being downloaded are even smaller.

The additional code complexity just doesn't seem justified.

igor-makarov:
This worries me:
> Is it possible for someone to delete and then redeploy a different version of a podspec. Is that an edge case that needs to be handled?
> Ex: I own pod A. I publish A/1.0.0. I realize I've made some mistake (Author name typo or something, not caught by lint). I delete podspec for A/1.0.0, fix my typo. Then do I need to publish it as A/1.0.1 (This seems ideal). Or can it be published again as A/1.0.0.
> Really hoping this isn't the case because that workflow woudn't work with these changes.

From my understanding, it's not only allowed, but also there's no way to determine such a case in the CDN builder, as any git operations on it will be impossibly long.

I know that most package registries do not allow deletions, in fact @dnkoutso was just as surprised as I was about CocoaPods allowing it 😂
@orta is there any particular reason deletions are allowed? Can we stop allowing them?

cltnschlosser:
> @orta is there any particular reason deletions are allowed? Can we stop allowing them?

If we can't, could trunk be updated to log deletions. And then block new podspecs with the same version as one that was deleted.

orta:
You can't redeploy a pod - no, the trunk DB keeps track of all versions and doesn't allow it

igor-makarov:
@orta this pod had deletion and re-addition of the same version:
https://github.com/CocoaPods/Specs/commits/master/Specs/5/f/7/CL_ShanYanSDK

orta:
Good spot 

Looks like we have [tests specifically for re-creating from a deletion](https://github.com/CocoaPods/trunk.cocoapods.org/blob/master/spec/functional/api/pods_controller_spec.rb#L192-L199) - I'm open to this changing

But probably we need to keep supporting deletions overall I think, at least making that decision is non-trivial and not just something I can offer oiff-hand

igor-makarov:
Regarding the current PR - I think this looks like a very corner-ish corner case as far as CDN is concerned, and also no simple solution comes to mind. 

I think we ought to move forward with this PR.

spec/cdn_source_spec.rb
```diff
@@ -190,6 +194,17 @@ def cleanup
         spec.name.should == 'BeaconKit'
         spec.version.should.to_s == '1.0.5'
       end
+
+      it 'doesn\'t attempt to access a version no in the version index' do
```
dnkoutso: Nit: not

igor-makarov:
Picked the nit

lib/cocoapods-core/cdn_source.rb
```diff
@@ -68,6 +69,21 @@ def preheat_existing_files
       end
     end
 
+    def files_definitely_to_update
+      [repo.join('**/*.yml'), repo.join('**/*.txt')].
```
dnkoutso: use `**/*.{yml,txt}`? This will invoke it once and could be faster and also no need to invoke `flatten`

lib/cocoapods-core/cdn_source.rb
```diff
@@ -68,6 +69,21 @@ def preheat_existing_files
...
+        map(&Pathname.method(:glob)).
+        flatten.
+        map { |f| f.relative_path_from(repo).to_s }
+    end
+
+    def deprecated_local_podspecs
+      download_file('deprecated_podspecs.txt')
+      relative_paths = local_file('deprecated_podspecs.txt', &:to_a).map(&:chomp).map(&Pathname.method(:new))
```
dnkoutso: combine into a single `map` call?
```ruby
local_file('deprecated_podspecs.txt', &:to_a).map { |p| Pathname.new(p.chomp) }
```

lib/cocoapods-core/cdn_source.rb
```diff
@@ -144,6 +160,10 @@ def versions(name)
     def specification_path(name, version)
       raise ArgumentError, 'No name' unless name
       raise ArgumentError, 'No version' unless version
+      unless versions(name).include?(Version.new(version))
+        raise StandardError, "Unable to find the specification #{name} " \
```
dnkoutso: nit of the nits wrap in backticks everything that is presented. Is `StandardError` or `Informative` here correct?

igor-makarov: I actually copied it from `Pod::Source` word-for-word. I'd rather keep it the same in both. Do you want to change it there too?

dnkoutso: ah oh well keep it as is!

igor-makarov: Cool!

igor-makarov:
I got rid of the extreme functional programming quirks as you requested 😂

spec/cdn_source_spec.rb
```diff
@@ -190,6 +194,17 @@ def cleanup
         spec.name.should == 'BeaconKit'
         spec.version.should.to_s == '1.0.5'
       end
+
+      it 'does not attempt to access a version no in the version index' do
```
dnkoutso: nit: "not"

dnkoutso:
just one build failure. feel free to merge once fixed!

spec/cdn_source_spec.rb
igor-makarov: Ah, I thought you meant replace `doesn\'t`.

igor-makarov:
Done!

igor-makarov:
Oh no, it still fails!

PR #562, opened 2019-07-13
[DSL] Support for test_type :ui

manuyavuz:
Add support for UI test type.

Required for CocoaPods/Cocoapods#9001.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ##### Enhancements
 
+* Add support for UI test specs with `test_type` value `:ui`  
+  [Yavuz Nuzumlali](https://github.com/manuyavuz)
+  [#562](https://github.com/CocoaPods/Core/pull/562)
+
 * Don't `Dir.chdir` when loading `Pod::Specification` from JSON  
   [Igor Makarov](https://github.com/igor-makarov)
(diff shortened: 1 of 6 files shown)


dnkoutso:
Nice! that easy huh? :)

Can you add a `CHANGELOG` entry and then a few specs in `dsl_spec`, `json_spec` and `linter_spec`? Those can mirror the tests for `test_type` `:unit`

manuyavuz:
Should we fail if developer does not provide `requires_app_host = true` along with `test_type = :ui`?

If so, I'm not sure if `linter.rb::_validate_test_type()` is the best place to do that. I don't want to add type specific if-else conditional in that method.

dnkoutso:
Seems like Xcode _does_ allow creating a UI test bundle without a target app. It fails to boot of course or gives a warning/error later so we can skip this check entirely as well.

Screen Shot 2019-07-13 at 3 23 48 PM

manuyavuz:
Yes it allows empty target and checks at runtime. Let's pass it, we don't need to be more protective than Xcode :)

dnkoutso:
Also let's expand  `attribute :test_type` documentation examples so the guides will be automatically updated it on next release.

dnkoutso:
Will need a rebase based on some last PRs that were merged.

PR #564, opened 2019-07-15
Fix CDNSource tests

igor-makarov:
I've fixed the flaky tests, ran it many times!

I've also found a small mistake in the repo update code. The repo update process ran `download_file('deprecated_podspecs.txt')` twice which was unnecessary.

### DIFF
--- a/spec/cdn_source_spec.rb
+++ b/spec/cdn_source_spec.rb
@@ -29,6 +29,11 @@ module Pod
         @path.join('.url').delete if @path.join('.url').exist?
       end
 
+      def print_dir(tag)
+        STDERR.puts tag
+        STDERR.puts Pathname.glob(@path.join('*')).sort.join("\n")
+      end
+
       @remote_dir = fixture('mock_cdn_repo_remote')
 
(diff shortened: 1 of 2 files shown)


igor-makarov:
@dnkoutso please review ASAP so that the main branch stops being red 😳

spec/cdn_source_spec.rb
```diff
@@ -29,6 +29,11 @@ def cleanup
         @path.join('.url').delete if @path.join('.url').exist?
       end
 
+      def print_dir(tag)
```
segiddins: this method appears tobe unused

igor-makarov: Yeah, but it's a useful one and it's in a spec. Is there a policy against helper code in tests?

segiddins: no, helper code is fine, but we shouldn't be commiting code that . never gets . run

igor-makarov: Well, where should one keep it for next time it's needed? Separate file?

PR #567, opened 2019-07-20
Update docs to mention that public header files should be included in source files

michaeleisel:
Please let me know if my understanding of this is correct! I was having issues because I had added a bunch of headers to `public_header_files` but not to `source_files`, and it wasn't working properly. Also, I wonder if this behavior should be changed somehow, even just adding a warning if it seems like the user is doing the wrong thing.

### DIFF
--- a/lib/cocoapods-core/specification/dsl.rb
+++ b/lib/cocoapods-core/specification/dsl.rb
@@ -1188,7 +1188,7 @@ module Pod
       #
       #   ---
       #
-      #   These patterns are matched against the source files to include headers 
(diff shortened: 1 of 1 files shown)


dnkoutso:
LGTM. need to run `bundle exec rubocop -a` to fix the rubocop issue or re-format the comment not to have trailing whitespace.

michaeleisel:
how do i merge?

dnkoutso:
You cannot unless you become a contributor. I can merge.

orta:
Thanks!

PR #568, opened 2019-07-22
[Consumer] Don't merge non-library spec build settings

segiddins:
That will now be handled by the xcconfig generator in CocoaPods

For https://github.com/CocoaPods/CocoaPods/pull/9030


dnkoutso:
branch seems out of date and is it possible to add a test here?

segiddins:
Branch is still mergable

PR #571, opened 2019-07-29
Added user_agent argument to http methods

abbeycode:
This way a caller, like the social media validator, can specify something different (Issue CocoaPods/CocoaPods#9049)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@
   [Eric Amorde](https://github.com/amorde)
   [CocoaPods/CocoaPods#8785](https://github.com/CocoaPods/CocoaPods/issues/8785)
 
-* Pass a non-browser user agent for social media validation
+* Pass a non-browser user agent for social media validation  
   [Dov Frankel](https://github.com/abbeycode)
   [#571](https://github.com/CocoaPods/Core/pull/571)
   [CocoaPods#9049](https://github.com/CocoaPods/CocoaPods/issues/9049)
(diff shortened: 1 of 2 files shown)


lib/cocoapods-core/http.rb
```diff
@@ -59,17 +59,19 @@ def self.validate_url(url)
     #
     # @return [REST::response]
     #
-    def self.perform_head_request(url)
+    def self.perform_head_request(url, user_agent)
       require 'rest'
 
-      resp = ::REST.head(url, 'User-Agent' => USER_AGENT)
+      user_agent ||= USER_AGENT
```
dnkoutso: can also set that as the default?
```ruby
def self.perform_head_request(url, user_agent = USER_AGENT)
...
end
```

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@
   [Eric Amorde](https://github.com/amorde)
   [CocoaPods/CocoaPods#8785](https://github.com/CocoaPods/CocoaPods/issues/8785)
 
+* Pass a non-browser user agent for social media validation
```
dnkoutso: nit add two empty spaces at the end of this line for markdown formatting.

lib/cocoapods-core/http.rb
```diff
@@ -8,11 +8,11 @@ module HTTP
     #
     # @return [string]
     #
-    def self.get_actual_url(url)
+    def self.get_actual_url(url, user_agent=nil)
```
dnkoutso: nit formatting is off here use `bundle exec rubocop -a`

lib/cocoapods-core/http.rb
abbeycode: I did at first, but realized this method is private to http.rb, so it becomes easier to pass nil at the call site(s) and then substitute it here.

PR #574, opened 2019-08-01
[DNM] Add the capability to analyze the Lockfile for presence of the old git spec repo

igor-makarov:
This adds two methods, one named `warn_if_contains_old_spec_repo` which is invoked when loading the `Lockfile` from the user directory.  
The second one, `remove_old_spec_repo` is meant to be used in very special cases and removes the git spec repo from the `Lockfile`.

### DIFF
--- a/lib/cocoapods-core/lockfile.rb
+++ b/lib/cocoapods-core/lockfile.rb
@@ -33,9 +33,12 @@ module Pod
     # @param  [Pathname] path
     #         the path where the lockfile is serialized.
     #
+    # @param  [Boolean] warn_if_contains_old_spec_repo
+    #         analyze the Lockfile and warn if it contains the old git spec repo.
+    #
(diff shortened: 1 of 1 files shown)


dnkoutso:
Hm this looks good but makes me wonder why have it "live" here? I guess ultimately as we've discussed between Core and CocoaPods it doesn't matter............

dnkoutso:
So lets say these steps happen:

- I use 1.7.5 today
- Have a `Podfile.lock` with some pods from master specs repo 
- I edit `Podfile` and remove `source` of the master specs repo _or_ replace it with `source 'https://cdn.cocoapods.org/'`
- I delete the git based master specs repo via `pod repo remove master`
- I run `pod install`

Does it all work and the `Podfile.lock` gets updated to specify `trunk` instead of `master`? Do I get a warning then?

PR #585, opened 2019-09-03
Add a Podfile option to reuse an external source specification

igor-makarov:
If a dependency was already defined using an external source, reuse that definition:
```ruby
    # define an external source
    target 'TargetA' do
      pod 'JSONKit', :podspec => 'https://example.com/JSONKit.podspec'
    end
    # reuse it
    target 'TargetB' do
      pod 'JSONKit', :use_previous_external_source => true
    end
```


segiddins:
Why does this need to be an option? It seems like we can default to this . behavior?

dnkoutso:
Agreed. further more this can also be a method inside the `Podfile` to simplify this?

igor-makarov:
@segiddins - Are we certain that this won't break any existing flow? 
@dnkoutso - Can you specify what kind of interface you're imagining?

dnkoutso:
```ruby
def jsonkit_pod
  pod 'JSONKit', :podspec => 'https://example.com/JSONKit.podspec'
end
```
And then use that method across targets?

igor-makarov:
That's what I'm using right now, but I thought a first-party way would be nicer for everyone.

I do think that @segiddins is right - it can be inferred and not need a separate option.   What do you think?

dnkoutso:
Inferring sounds good.

igor-makarov:
Tried to hack it today, this seems non trivial.

I'll try to find time for it on the weekend.

igor-makarov:
Done, figured it out!

Some of the tests were failing because iterating over dependencies mutated the target definitions.   
I've changed some of the getters on `TargetDefinition` to not add a `nil` value to the hash in the getter. This didn't break anything AFAIK but please review it thoroughly.

lib/cocoapods-core/podfile/dsl.rb
```diff
@@ -273,6 +286,18 @@ def pod(name = nil, *requirements)
           raise StandardError, 'A dependency requires a name.'
         end
 
+        if requirements.empty? || requirements.last.is_a?(Hash)
+          theoretical_dependency = Dependency.new(name, requirements.last)
+          if theoretical_dependency.external_source.nil?
+            previously_defined_dependency = dependencies.find do |dependency|
```
segiddins: how will this linear scan perform on very large podfiled?

lib/cocoapods-core/podfile/target_definition.rb
```diff
@@ -174,7 +174,7 @@ def abstract=(abstract)
       # @return [String] the inheritance mode for this target definition.
       #
       def inheritance
-        get_hash_value('inheritance', 'complete')
+        get_hash_value('inheritance') || 'complete'
```
segiddins: why these changes?

igor-makarov: I've changed several getters to avoid mutating the hash because iterating over deps mutated them. 

The code assumed, previously, (and implicitly) that deps only get evaluated after the `Podfile` is completely scanned.

However, going over previously added deps before the entire `Podfile` is scanned mutated things and now it doesn't.

lib/cocoapods-core/podfile/dsl.rb
igor-makarov: There's a bit of a Shlemiel effect here, but ultimately it's all done in-memory and my profiling for 70 total pods shows an increase from 3% samples to 7%. 

What do you think? Should I optimize now or can this be done later? I mean, I've made several optimizations recently that outweigh this by a margin 😂

lib/cocoapods-core/podfile/target_definition.rb
segiddins: :+1:. Is there a test that will fail if we accidentally go back to the old behavior later?

lib/cocoapods-core/podfile/dsl.rb
segiddins: Probably fine for now, but we might want to eventually put it in a hash for constant-time lookup

lib/cocoapods-core/podfile/target_definition.rb
dnkoutso: want to continue this @igor-makarov ?

igor-makarov: Hadn't had the time to think about this yet.. Sorry 😭

PR #588, opened 2019-09-18
Added work with netrc file to perform autologin if CDN repo requires basic auth

igorudovika:
Fix  for https://github.com/CocoaPods/CocoaPods/issues/9151 issue

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ group :development do
   gem 'bacon'
   gem 'mocha'
   gem 'mocha-on-bacon'
+  gem 'netrc'
   gem 'prettybacon'
   gem 'rake'
   gem 'vcr'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -60,6 +60,7 @@ GEM
       mocha (>= 0.13.0)
     multi_json (1.12.1)
     nap (1.1.0)
+    netrc (0.11.0)
     notify (0.5.2)
     parser (2.4.0.0)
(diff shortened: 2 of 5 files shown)


dnkoutso:
cc @igor-makarov want to take a stab at reviewing this?

igor-makarov:
I'm thinking that #580 would allow for a much more straightforward implementation, as cURL and Typhoeus (which wraps it) support `.netrc` out of the box.

igorudovika:
> cURL
I have done few tests with Typhoeus. Unfortunately it does not work. Seems credentials from .netrc is taken into request only for command line `curl` version. Same true for .curlrc

igor-makarov:
Did yoy check this option out?
https://github.com/typhoeus/ethon/blob/cf1d26158755de7a536ac8da3d344ae567bea719/lib/ethon/curls/options.rb#L305

igorudovika:
> Did you check this option out?
> https://github.com/typhoeus/ethon/blob/cf1d26158755de7a536ac8da3d344ae567bea719/lib/ethon/curls/options.rb#L305

nope, let me check

igorudovika:
> > Did you check this option out?
> > https://github.com/typhoeus/ethon/blob/cf1d26158755de7a536ac8da3d344ae567bea719/lib/ethon/curls/options.rb#L305
> 
> nope, let me check

yep, with such option it works. Ok then I will wait #580 merge and update PR after

PR #595, opened 2019-10-02
Ensure that specs can safely round-trip through JSON

segiddins:


dnkoutso:
Can we target `1-8-stable` for this so we can ship it with 1.8.2?

justinseanmartin:
The alternative that we'd discussed was we could create a new hash and copy the keys/values over into the new hash. I think that would make the intent more clear, and be more future proof. We could do that change later though.

PR #602, opened 2019-10-24
Make initializing `Platform` with a string 'macos' behave the same as `Platform.macos`

amorde:
This makes it easier for `Platform` consumers to use `Platform.new` to parse values from JSON, Info.plist files,  etc. without having to manually map `macos` => `osx` in order for it to successfully match other platforms defined within CocoaPods.

### DIFF
--- a/spec/platform_spec.rb
+++ b/spec/platform_spec.rb
@@ -29,8 +29,14 @@ module Pod
       end
 
       it 'can be initialized with a string symbolic name' do
-        Platform.new('ios')
-        @platform.name.should == :ios
+        platform = Platform.new('ios')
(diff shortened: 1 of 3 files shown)


spec/platform_spec.rb
```diff
@@ -29,8 +29,14 @@ module Pod
       end
 
       it 'can be initialized with a string symbolic name' do
-        Platform.new('ios')
-        @platform.name.should == :ios
+        platform = Platform.new('ios')
```
amorde: this was initializing a new platform and throwing it away, and then checking state initialized elsewhere. not an effective test

PR #612, opened 2020-01-17
[Pod::Version prerelease?] Not correct when version with metadata.

sagiwei:
According to the [semver](https://semver.org/), metadata could contain ASCII alphanumerics. That make `Pod::Version` method `prerelease?` regard version like '1.1.0+exp.1' as a prerelease version.


dnkoutso:
@sagiwei can you rebase?

sagiwei:
@dnkoutso ，yeah, done.

PR #626, opened 2020-04-25
Add a lint check for the name format of vendored static libraries

amorde:
This checks that vendored libraries match the expected format `lib[name].a`.
This comes up occasionally in issues, such as https://github.com/CocoaPods/CocoaPods/issues/9596

### DIFF
--- a/lib/cocoapods-core/specification/linter.rb
+++ b/lib/cocoapods-core/specification/linter.rb
@@ -337,6 +337,19 @@ module Pod
         end
       end
 
+      # Performs validations related to the `vendored_libraries` attribute.
+      #
+      # @param [Array<String>] vendored_libraries the values specified in the `vendored_libraries` attribute
+      #
+      def _validate_vendored_libraries(vendored_libraries)
+        vendored_libraries.each do |lib|
+          lib_name = lib.downcase
+          unless lib_name.end_with?('.a') && lib_name.start_with?('lib')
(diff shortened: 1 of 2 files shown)


IvanRomanovski:
@amorde  Is this meant to be a breaking change? 

My previously working podspec with:
```
s.vendored_libraries = '**/*.a'
```
now fails with
```
WARN  | [iOS] vendored_libraries: `*.a` does not match the expected static library name format `lib[name].a`
```
It seems that this validation understands "*" as a literal instead of any value.

lib/cocoapods-core/specification/linter.rb
```diff
@@ -337,6 +337,19 @@ def _validate_libraries(libs)
...
       end
 
+      # Performs validations related to the `vendored_libraries` attribute.
+      #
+      # @param [Array<String>] vendored_libraries the values specified in the `vendored_libraries` attribute
+      #
+      def _validate_vendored_libraries(vendored_libraries)
+        vendored_libraries.each do |lib|
+          lib_name = lib.downcase
+          unless lib_name.end_with?('.a') && lib_name.start_with?('lib')
```
IvanRomanovski, line 347: @amorde it seems like you are checking full file name (with path) instead of just basename like in the line below.

dnkoutso:
@IvanRomanovski we can improve and fix this. It seems like indeed the glob must be expanded first? Want to take a stab at it and we can ship a fix in 1.10.1?

PR #636, opened 2020-06-26
Raise macOS default deployment target

DarkDust:
Raise the default deployment target for macOS from 10.6 to 10.9. This gets rid of a warning in Xcode 12 with pods that did not explicitly specify a deployment target.

Closes CocoaPods/CocoaPods#9884

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,9 @@
 
 ##### Bug Fixes
 
-* None.  
-
+* Raise macOS default deployment target to avoid a warning in Xcode 12
+  [Marc Haisenko](https://github.com/DarkDust)
+  [CocoaPods#9884](https://github.com/CocoaPods/CocoaPods/issues/9884)
 
 ## 1.9.3 (2020-05-29)
 
--- a/lib/cocoapods-core/podfile/dsl.rb
+++ b/lib/cocoapods-core/podfile/dsl.rb
@@ -566,7 +566,7 @@ module Pod
       # Specifies the platform for which a static library should be built.
       #
       # CocoaPods provides a default deployment target if one is not specified.
-      # The current default values are `4.3` for iOS, `10.6` for OS X, `9.0` for tvOS
+      # The current default values are `4.3` for iOS, `10.9` for OS X, `9.0` for tvOS
       # and `2.0` for watchOS.
       #
       # If the deployment target requires it (iOS < `4.3`), `armv6`
--- a/lib/cocoapods-core/podfile/target_definition.rb
+++ b/lib/cocoapods-core/podfile/target_definition.rb
@@ -611,7 +611,7 @@ module Pod
 
       #--------------------------------------#
 
-      PLATFORM_DEFAULTS = { :ios => '4.3', :osx => '10.6', :tvos => '9.0', :watchos => '2.0' }.freeze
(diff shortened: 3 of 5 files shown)


segiddins:
This is a breaking change, since it will dirty everyone’s Podfile.lock that contains a pod that is using the default

DarkDust:
What does that mean for me? Is there some documentation or other action by me required, or do I just need to wait for other devs/maintainers to chime in?

dnkoutso:
@DarkDust it probably means we cant merge it until we ship a 2.0 version of CocoaPods. For your original issue just supply the version of macOS in your `Podfile` to fix the Xcode warning.

DarkDust:
I did specify `platform :osx, '10.9'` in my `Podfile`, but that doesn't change the deployment target version of the pods that specify a platform but omit the version. Well, it's just a harmless warning so not that important. It can wait.

alexito4:
Just FYI this is also happening with iOS projects since iOS 8.0 is no longer supported. What's the solution here? It's expected that Pods update and specify a version instead of relying on the defaults?

Kaspik:
@dnkoutso @segiddins I'm not sure about the requirement "wait for 2.0.0 Cocoapods version" here.

CocoaPods currently set `4.3` for iOS or `10.6` for OS  (https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/podfile/dsl.rb#L569) - and if libraries are not providing the info, they will set these values. But these values make projects non-compilable. So *everyone* needs to add `post-install` hook to make their project compile.

If you would update these default values to `9.0` for iOS and 10.9 for macOS with next release of CocoaPods 1.11.0, what exactly is the breaking change?

- everyone with post hook won't see any difference
- everyone with library that specifies this won't see any difference
- everyone with library that doesn't specify this and without post hook... is probably using older CocoaPods anyways (so can't upgrade anyways)

So current default values don't work and don't make project compilable, right?
Can we please do (not only) this update and update default values that were set *more than 6 years ago*?

Thanks

paulb777:
@Kaspik See discussions at https://github.com/CocoaPods/CocoaPods/pull/10070 and https://github.com/CocoaPods/CocoaPods/issues/9884

tldr: podspecs and Podfiles should be updated

Kaspik:
I saw it (I even was part of the convo: https://github.com/CocoaPods/CocoaPods/issues/9884#issuecomment-651600153), but that doesn't change anything on my comment, right?

Either CocoaPods should not have `any` default value, and leave it up to post install hook / libraries as you suggest, or it should offer `valuable` default value. Right now it doesn't check any of these two checkboxes...

Another option would be set the `minimum target` based on `platform` `Podfile` definition, not `default value` from `CocoaPods` `Core`. But that's a bigger change than just bumping the minimum values.

paulb777:
@Kaspik Sorry, on a closer read, I see your point, I was answering about a Podfile default versus a podspec default.

The potential breakage with changing the podspec default is much more theoretical - a pod using an iOS 8 API that is deprecated or deleted in iOS 9 and consumers still using Xcode 11.  However, it would still broken either way with Xcode 12.

Kaspik:
Hmmm, yeah, theoretically yes... Is there any reason why not to come up with 2.0.0 instead of 1.11.0 then? Or what's the timeline for CocoaPods for upcoming months? This seems like a reasonable update sooner rather than later.

PR #649, opened 2020-09-19
Support sparse checkouts in git Source

igor-makarov:


lib/cocoapods-core/source.rb
```diff
@@ -464,6 +466,38 @@ def unchanged_github_repo?
...
+      relative_pod_path = path.relative_path_from(repo)
+
+      unless sparse_checkout_file.include?(relative_pod_path.to_s)
+        File.open(sparse_checkout_config_path, 'a') do |sparse|
+          sparse.puts relative_pod_path
+        end
+        @sparse_checkout_file = nil
+      end
+
+      STDERR.puts relative_pod_path
```
dnkoutso, line 497: nit, is this debugging code?

igor-makarov: whoops - fixed...

lib/cocoapods-core/source.rb
```diff
@@ -464,6 +466,49 @@ def unchanged_github_repo?
...
+    end
+
+    def sparse_checkout?
+      if @is_sparse_checkout.nil?
+        @is_sparse_checkout = File.exist?(sparse_checkout_config_path)
+      end
+      @is_sparse_checkout
+    end
+
+    def sparse_checkout_file
```
segiddins, line 483: nit: this method doesn't return a file, but rather a list of strings, the name should reflect that

lib/cocoapods-core/source.rb
```diff
@@ -111,7 +111,9 @@ def specs_dir
     #         stored.
     #
     def pod_path(name)
-      specs_dir.join(*metadata.path_fragment(name))
+      path = specs_dir.join(*metadata.path_fragment(name))
+      add_to_sparse_checkout(path) if sparse_checkout?
```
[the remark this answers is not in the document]
igor-makarov: Well, it helps to view it as a "lazy" repo - querying for a path simply ensures that the path can be queried correctly.  
If viewed this way, it's not a "side effect" but a necessary implementation detail.

Another way to do this is to add the call to `#add_to_sparse_checkout` everywhere the path is being queried, but that increases the risk of bugs. My proposition seems more robust to me.

lib/cocoapods-core/source.rb
igor-makarov: done!

PR #651, opened 2020-09-28
Add GitHub Actions Specs workflow

igor-makarov:
Travis is slow, this is a full Core test build running in GH Actions.


dnkoutso:
@igor-makarov can you point to `1-10-stable`?

igor-makarov:
@dnkoutso Done!

.github/workflows/Specs.yml
```diff
@@ -0,0 +1,48 @@
...
+  specs:
+    strategy:
+      matrix:
+        os: [ubuntu-16.04]
+        ruby: [2.3.3, 2.4, 2.5, 2.6, 2.7]
+        include:
+          - os: macos-10.15
+            ruby: system
+
+    name: Ruby ${{ matrix.ruby }}
```
endocrimes, line 13: Can you include the OS in the workflow name? Makes it much easier to skim

.github/workflows/Specs.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          gem install bundler -v "~> 1.17"
+          bundle install --jobs 4 --retry 3 --without debugging documentation
+
+      - name: Run Specs + Rubocop
+        run: bundle exec rake spec
+
+on:
+  push:
+    branches:
```
endocrimes, line 41: Is there a reason to not run on all branches? (Getting a ci run before opening the PR itself can be nice sometimes)

endocrimes (commented):
Thanks for doing this!

.github/workflows/Specs.yml
igor-makarov: @endocrimes - I can't include just the OS type, and I figured that saying `ubuntu-16.04` is a bit verbose. Do you think it's better to include it, even if verbose?

.github/workflows/Specs.yml
igor-makarov: @endocrimes It used to be the case that this caused a weird duplication in test runs that would show up in checks. I'll try and see if it's still the case.

igor-makarov: @endocrimes Yup, it did that. Reverting 😭

.github/workflows/Specs.yml
igor-makarov: @endocrimes I have added the runner name to the job title - looks a bit verbose (e.g. `Specs / ubuntu-16.04 / Ruby 2.6 (pull_request)`) but livable.

endocrimes: Verbose is good for things like this I think 😅  - Thanks for adding it!

PR #657, opened 2020-10-22
[Linter] Allow version `0` to be used.

alloy:
In React and React Native, nightly builds are created using a version like `0.0.0-githash`. Trying to install such versions with CocoaPods currently fails due to this linter check.

I wasn't able to find a rationale for not allowing version `0` nor could I come up with a reason for not allowing it.

See:
* https://github.com/microsoft/react-native-macos/issues/620
* https://github.com/facebook/react-native/issues/30036

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Allow version `0` to be used.  
+  [Eloy Durán](https://github.com/alloy)
+  [#657](https://github.com/CocoaPods/Core/pull/657)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Allow version `0` to be used.
```
endocrimes, line 7: nit: missing double space 😅

endocrimes (approved):
This seems fine to me. They'd be weird if they were pushed to the specs repo, but this is a common enough pattern for nightly things that we should probably allow it

CHANGELOG.md
alloy: Done

dnkoutso (approved):

brunolemos:
Could we have a 1.11.0 RC or similar? Thank you.

alloy:
@brunolemos I assume you’re asking because it allows nightly builds of React Native to be installed? (Just adding some context.)

brunolemos:
yes :)

PR #662, opened 2020-11-25
Fix typo

friederbluemle:
Minor typo fix: `perfomed` -> `performed`

as seen in this message:

```
Resolving dependencies of `Podfile`
  CDN: trunk Relative path: CocoaPods-version.yml exists! Returning local because checking is only perfomed in repo update
```


dnkoutso:
thanks!

PR #663, opened 2020-11-26
Ignore path when checking vendored_libraries  format.

IvanRomanovski:
Lint check for the name format of vendored static libraries (https://github.com/CocoaPods/Core/pull/626) introduced a breaking change, where relative paths and globs like `**/*.a` are not supported anymore.

This PR modifies linter to allow relative paths.  Globs like `**/*.a` still can't be supported because this would break the requirement that static libraries should begin with *lib*, however this limitation can be easily overcome by modifying `**/*.a` to `**/lib*.a` in a podspec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
-* None.  
+* Ignore path when checking vendored_libraries `lib[name].a` format.
+  [Ivan Romanovski](https://github.com/IvanRomanovski)
+  [#626](https://github.com/CocoaPods/Core/pull/626)
 
 
 ## 1.10.0 (2020-10-20)
--- a/lib/cocoapods-core/specification/linter.rb
+++ b/lib/cocoapods-core/specification/linter.rb
@@ -340,9 +340,10 @@ module Pod
       #
       def _validate_vendored_libraries(vendored_libraries)
         vendored_libraries.each do |lib|
-          lib_name = lib.downcase
+          basename = File.basename(lib)
+          lib_name = basename.downcase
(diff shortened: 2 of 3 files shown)


dnkoutso:
@IvanRomanovski this is a good change but we think we can actually validate the values _after_ the vendored_libraries glob is expanded which happens in the CocoaPods project in the validator.rb class instead of here.

amorde (approved):
Thanks! I think the long term path is to move this to the validator that lives within CocoaPods instead of keeping it here in Core, so we can use the result of computing the glob in the validation. But this is an incremental improvement 👍

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
-* None.  
+* Ignore path when checking vendored_libraries `lib[name].a` format.
```
amorde, line 13: This is a bit misleading, maybe change to `support **/lib*.a in vendored_libraries when linting`

amorde: Ah I see what you mean by ignoring the path

segiddins (approved):

dnkoutso:
Going to implement what @amorde said here https://github.com/CocoaPods/Core/pull/663#pullrequestreview-555092476 instead for 1.11.0.

Will close this PR for now and open new ones.

dnkoutso:
Re-implemented it in https://github.com/CocoaPods/Core/pull/693 and https://github.com/CocoaPods/CocoaPods/pull/10832 in order to be able to use globs again in `vendored_libraries` and also catch the misnamed libraries.

PR #639, opened 2020-07-18
Replace Typhoeus with a native HTTP/2 client

igor-makarov:

PR #669, opened 2021-01-11
Replace Typhoeus with `async-http`

jasl:
`typhoeus` is the only blocking problem that preventing CocoaPods running on M1 natively

I have read many GH issues of`ffi` and `ethon`, it looks like the segfault problem is hard to fix, so I decide to rewrite it.

I'm using [async-http](https://github.com/socketry/async-http) to replace `typhoeus` + `concurrent-ruby`

Advantage:

- can running on native M1 Ruby
- the abstraction is simpler than `typhoeus` + `concurrent-ruby`
- the author ioquatix is also the author of Ruby 3 Fiber scheduler

Disadvange:

- Minimal Ruby version require 2.5+

Once blocking PRs getting merged, I'll cleanup commits and ready for reviewing


jasl:
image

It works on my M1 mac, but lacking

- Complete codes
- Update tests

amorde:
Thanks for the PR! this has been top of mind recently with the M1 Mac releases.

There's an existing PR that attempts to achieve the same result here https://github.com/CocoaPods/Core/pull/639 - I'm not not familiar enough with either of these APIs to pick one, but I'd love to move at least one of these PRs forward. @igor-makarov what do you think?

igor-makarov:
First of all, thanks for all the work you've put in, @jasl!  
I've got several comments on your proposed change:

* Bumping Ruby/Rubocop version should probably be done as a separate PR. It seems like a big deal.
* `async` relies on another gem with native extensions, `nio4r`. Does it work on M1?
* GitHub Actions doesn't have Apple Silicon machines yet. Official support for native M1 should probably wait until we can actually run tests on it.
* `typhoeus` is also used in the [CocoaPods repo](https://github.com/CocoaPods/CocoaPods/blob/35be8e23d3baebf07098f95da041b3999901856a/lib/cocoapods/sources_manager.rb#L74). I think the `cdn_url?` function can be moved to Core, but this will involve integration work in any case.
* I've tried running basic operations `install`, `repo update` and I've noticed that it fails on responses that have more than one chunk. I think that `async-http` returns the response without the body being complete, see [here](https://github.com/socketry/async-http/blob/master/examples/stream/stop.rb) for an example of collecting chunks.
* Considering the previous item, may I suggest using the already-built concurrency and synchronization and just change the implementation of `download_typhoeus_impl_async` to use `async-http` and resolve `Future`s with it? 
* There's going to be a lot of work getting the tests necessary for this transition.

jasl:
> First of all, thanks for all the work you've put in, @jasl!
> I've got several comments on your proposed change:
> 
> * Bumping Ruby/Rubocop version should probably be done as a separate PR. It seems like a big deal.
> * `async` relies on another gem with native extensions, `nio4r`. Does it work on M1?
> * GitHub Actions doesn't have Apple Silicon machines yet. Official support for native M1 should probably wait until we can actually run tests on it.
> * `typhoeus` is also used in the [CocoaPods repo](https://github.com/CocoaPods/CocoaPods/blob/35be8e23d3baebf07098f95da041b3999901856a/lib/cocoapods/sources_manager.rb#L74). I think the `cdn_url?` function can be moved to Core, but this will involve integration work in any case.
> * I've tried running basic operations `install`, `repo update` and I've noticed that it fails on responses that have more than one chunk. I think that `async-http` returns the response without the body being complete, see [here](https://github.com/socketry/async-http/blob/master/examples/stream/stop.rb) for an example of collecting chunks.
> * Considering the previous item, may I suggest using the already-built concurrency and synchronization and just change the implementation of `download_typhoeus_impl_async` to use `async-http` and resolve `Future`s with it?
> * There's going to be a lot of work getting the tests necessary for this transition.

> Bumping Ruby/Rubocop version should probably be done as a separate PR. It seems like a big deal.

Yes, `async` require at least Ruby 2.5.0, depends on CocoaPods drop support of 10.14 and below,
that's should be decide by the Core team.

but for now, I add these changes in the branch for not blocking my work.

> async relies on another gem with native extensions, nio4r. Does it work on M1?

It works, my branch can natively running on M1 (see the screenshot above).
Besides `nio4r` has a pure Ruby implementation ([but not enable in M1](https://github.com/socketry/nio4r/blob/master/lib/nio.rb#L17))

> GitHub Actions doesn't have Apple Silicon machines yet. Official support for native M1 should probably wait until we can actually run tests on it.

For now, I can only run tests on my M1 Mac, Homebrew also suffer this,
but I think this is fine, if CI green, most probably no issue on M1...

> typhoeus is also used in the CocoaPods repo. I think the cdn_url? function can be moved to Core, but this will involve integration work in any case.

I've sending another PR https://github.com/CocoaPods/CocoaPods/pull/10346

> I've tried running basic operations install, repo update and I've noticed that it fails on responses that have more than one chunk. I think that async-http returns the response without the body being complete, see here for an example of collecting chunks.

I'll take a look, the PR not ready yet, hopefully I can complete this in this or next week.

> Considering the previous item, may I suggest using the already-built concurrency and synchronization and just change the implementation of download_typhoeus_impl_async to use async-http and resolve Futures with it?

I can try, but I don't prefer this idea,

because:

- `aysnc-http` is combine of `async` + `http` which is competitive with `concurrent-ruby` + `typhoeus`
  - if we have to keeping `concurrent-ruby`, your branch is better than me in my view
- Ruby 3 has built-in non-blocking solution called Fiber scheduler, `async` already use that (they are the same author)
  - I believe `nio4r` will be removed for Ruby 3.0 implementation
  - I heard macOS will not bundle Ruby in future, when the times come, Cocoapods can migrate to Ruby 3.0

> There's going to be a lot of work getting the tests necessary for this transition.

Agree, made Cocoapods running on M1 natively is a long-term work, but we can learn from Homebrew:

- Make the codebase compatible with M1 in theroy
- Ensure not break x86 platform
- Let M1 users test in their own risk

jasl:
> Considering the previous item, may I suggest using the already-built concurrency and synchronization and just change the implementation of download_typhoeus_impl_async to use async-http and resolve Futures with it?

@amorde could you consider this suggest? I left some opinion in https://github.com/CocoaPods/Core/pull/669#issuecomment-758575513

but if it worthy keeping `concurrent-ruby`, I prefer https://github.com/CocoaPods/Core/pull/639

igor-makarov:
@jasl We aren't tied to Concurrent-Ruby - it's an implementation detail of CDNSource. 

The function it serves in the current implementation is structuring the concurrent HTTP/2 requests so that as many as possible can be performed in parallel. 

For what I gather, `async` provides a different way to structure this concurrency - so my guess is that the whole implementation needs to be rewritten.

I'm not very familiar with this type of concurrency so I don't know how to structure it correctly. However, the passing of `context` seems like it could be factored out somehow.  
In the original implementation, all requests run on the `libcurl` reactor - even when there is only one.

igor-makarov:
@jasl I've managed to fix the partial HTTP body issue without hurting concurrency. I've pushed the change to your branch.

I'm repeating the list from above to focus the discussion:

* Bumping Ruby/Rubocop version should probably be done as a separate PR. It seems like a big deal.
* `async` relies on another gem with native extensions, `nio4r`. Does it work on M1?
* GitHub Actions doesn't have Apple Silicon machines yet. Official support for native M1 should probably wait until we can actually run tests on it.
* `typhoeus` is also used in the [CocoaPods repo](https://github.com/CocoaPods/CocoaPods/blob/35be8e23d3baebf07098f95da041b3999901856a/lib/cocoapods/sources_manager.rb#L74). I think the `cdn_url?` function **should** be moved to Core.
* There's going to be a lot of work getting the tests necessary for this transition.

jasl:
> @jasl I've managed to fix the partial HTTP body issue without hurting concurrency. I've pushed the change to your branch.
> 
> I'm repeating the list from above to focus the discussion:
> 
> * Bumping Ruby/Rubocop version should probably be done as a separate PR. It seems like a big deal.
> * `async` relies on another gem with native extensions, `nio4r`. Does it work on M1?
> * GitHub Actions doesn't have Apple Silicon machines yet. Official support for native M1 should probably wait until we can actually run tests on it.
> * `typhoeus` is also used in the [CocoaPods repo](https://github.com/CocoaPods/CocoaPods/blob/35be8e23d3baebf07098f95da041b3999901856a/lib/cocoapods/sources_manager.rb#L74). I think the `cdn_url?` function **should** be moved to Core.
> * There's going to be a lot of work getting the tests necessary for this transition.

Thank you! I'll check tomorrow

> Bumping Ruby/Rubocop version should probably be done as a separate PR. It seems like a big deal.

Will separate another PR for this before this one ready to review

BTW. I've sent https://github.com/CocoaPods/Core/pull/670 for remove forked json gem

> async relies on another gem with native extensions, nio4r. Does it work on M1?

It works, I've tested in my M1 Mac Mini, I'll do double check

> GitHub Actions doesn't have Apple Silicon machines yet. Official support for native M1 should probably wait until we can actually run tests on it.

We should ensure all exist targets pass, I'll also run test on my M1 Mac

> typhoeus is also used in the CocoaPods repo. I think the cdn_url? function should be moved to Core.

A PR have sent see https://github.com/CocoaPods/CocoaPods/pull/10346

igor-makarov:
@jasl I've reviewed https://github.com/CocoaPods/CocoaPods/pull/10347.

I'll be waiting for the other items for now.

jasl:
> @jasl I've reviewed [CocoaPods/CocoaPods#10347](https://github.com/CocoaPods/CocoaPods/pull/10347).
> 
> I'll be waiting for the other items for now.

Thank you, when https://github.com/CocoaPods/Core/pull/670 merged, I'll separate a PR including:

- Require minimal Ruby to 2.5
  - newer Rubocop no longer support Ruby 2.3
  - async require Ruby 2.5+
  - 2.5 is an important version of Ruby
- Upgrade to Rubocop to 1.8
  - 0.38 too old that made my RubyMine unhappy

Important: if this PR getting merge means CocoaPods no longer support macOS 10.14 system Ruby (2.3.7 if my memory correct)

igor-makarov:
Ok, so to review (each item ideally should be separate PR to review):
1. #670
2. After that is taken care of, create a PR with just the Ruby version deprecation.
3. Typhoeus removal PR in Core would consist at this point of only Typhoeus removal.
4. Because https://github.com/CocoaPods/CocoaPods gem is pinned to Core, deprecate Ruby version there as well and update the version pin.
5. Update the version pin of https://github.com/CocoaPods/CocoaPods again, this time removing Typhoeus transitive dependency.

jasl:
> Ok, so to review (each item ideally should be separate PR to review):
> 
> 1. #670
> 2. After that is taken care of, create a PR with just the Ruby version deprecation.
> 3. Typhoeus removal PR in Core would consist at this point of only Typhoeus removal.
> 4. Because https://github.com/CocoaPods/CocoaPods gem is pinned to Core, deprecate Ruby version there as well and update the version pin.
> 5. Update the version pin of https://github.com/CocoaPods/CocoaPods again, this time removing Typhoeus transitive dependency.

no problem

amorde:
I think we can bump Ruby to 2.6 which is shipped with 10.15. If we're going to drop support for 10.14 system ruby, we might as well update to 10.15 system Ruby and use the newest version possible.

jasl:
> I think we can bump Ruby to 2.6 which is shipped with 10.15. If we're going to drop support for 10.14 system ruby, we might as well update to 10.15 system Ruby and use the newest version possible.

I'm hoping this, I'd like to do some contributions to help CocoaPods future-proof!

igor-makarov:
@jasl @amorde agreed about 10.15 - we aren't even testing on 10.14 because GH doesn't have the runners for it.

jasl:
Since https://github.com/CocoaPods/CocoaPods/pull/10346 nearly ready

I'll back to work on this PR on weekend,
I plan to refactor `download_file_async` and `download_impl_async` resolving propagating issue first (current impl can't rescue error such as Timeout, NetworkError, etc. in async context)
then update and fix tests

igor-makarov:
@jasl could you start with creating the Ruby deprecation PR? It will get the ball rolling and begin getting reviewed.

jasl:
> @jasl could you start with creating the Ruby deprecation PR? It will get the ball rolling and begin getting reviewed.
> 
> Minimum Ruby version = 2.6.

No problem, one more question, can I upgrade bundler in that PR? since Ruby 2.6 bundled bundler 2.0, and 1.8.7 won't compatible with 2.0's Gemfile.lock

igor-makarov:
@jasl not sure I understand, what isn't compatible with current Bundler `1.17.3`? BTW, that's just the version for installing on CI, I believe users can use other versions.

jasl:
> @jasl not sure I understand, what isn't compatible with current Bundler `1.17.3`? BTW, that's just the version for installing on CI, I believe users can use other versions.

This change not affect users, they install CocoaPods via Rubygems.

I see https://github.com/CocoaPods/CocoaPods/pull/10324 also upgrade Bundler.

Bundler only be used in developer and CI, 
upgrade to 2.x means CocoaPods developers and CI no longer need  `gem install bundler -v "~> 1.17"`.

Besides, Bundler 1.17 has deprecation warning on Ruby 3.0

```
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:29: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:35: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:35: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:35: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:44: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:35: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:44: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
/Users/jasl/.rvm/gems/ruby-3.0.0/gems/bundler-1.17.3/lib/bundler/shared_helpers.rb:118: warning: Pathname#untaint is deprecated and will be removed in Ruby 3.2.
```

igor-makarov:
@jasl I think you're right. Also, it seems that going to Ruby 2.6 can allow us to finally get rid of the janky `activesupport` version pin:
```
  # 6.0 requires Ruby 2.5.0
  s.add_runtime_dependency 'activesupport', '> 5.0', '< 6'
```

jasl:
> @jasl I think you're right. Also, it seems that going to Ruby 2.6 can allow us to finally get rid of the janky `activesupport` version pin:
> 
> ```
>   # 6.0 requires Ruby 2.5.0
>   s.add_runtime_dependency 'activesupport', '> 5.0', '< 6'
> ```

Yes, I believe AS 5.x has problem on Ruby 2.7+

igor-makarov:
Cool, so waiting for a PR with:
* Ruby minimum 2.6.0
* Bundler bump (CI only)
* Rubocop bump (dev only)
* ActiveSupport version specification bumped to '~> 6'

jasl:
> Cool, so waiting for a PR with:
> 
> * Ruby minimum 2.6.0
> * Bundler bump (CI only)
> * Rubocop bump (dev only)
> * ActiveSupport version specification bumped to '~> 6'

Correct: Bundler bump (CI & Dev only)

will do later

igor-makarov:
Thanks for all the work!

jasl:
Yay!
I've pushed the first all green commits!

I've squashed into one commit https://github.com/CocoaPods/Core/pull/669/commits/c30e8eb3f02012599305e5e3c459b6286bb4f73d

The codes looks ugly, and I may misunderstand original intentions

Please give me some requirements, advices, directions, style suggestions, etc. then I'll do a polishment.

jasl:
notes:

- `webmock` doesn't support HTTP2, so I force use HTTP1 in test
- `async-http` `response.read` will read all chucks, see the usage in https://github.com/socketry/async-http/issues/37#issuecomment-560189660

jasl:
It's weird in my local specify `Async::HTTP::Protocol::HTTP1` no problem
but in CI it raise error deep in `async-http`... maybe a webmock bug

jasl:
running tests in my M1 mac

Must use `NIO4R_PURE=true` to enable pure Ruby implementaion on system Ruby,
no need for custom build

`NIO4R_PURE=true bundle exec rake spec`

```
1280 specifications (2275 requirements), 0 failures, 0 errors

--------------------------------------------------------------------------------
Checking code style...
--------------------------------------------------------------------------------

warning: parser/current is loading parser/ruby26, which recognizes
warning: 2.6.6-compliant syntax, but you are running 2.6.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Running RuboCop...
Inspecting 43 files
...........................................

43 files inspected, no offenses detected
jasl@bogon:~/Workspaces/Ruby/Core on async-http$ ruby -v
ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.arm64e-darwin20]
jasl@bogon:~/Workspaces/Ruby/Core on async-http$ NIO4R_PURE=true bundle exec rake spec
```

jasl:
Rebase into https://github.com/CocoaPods/Core/pull/671

I think my first iteration is over, please review

jasl:
Just log a weird thing

when first run test on my M1 Mac (no matter what Ruby) it fails with

```
Bacon::Error: "file system".==("git") failed
	spec/source_spec.rb:18:in `block (3 levels) in <module:Pod>': In general - return its type when git
	/Users/jasl/Workspaces/Ruby/Core/spec/spec_helper.rb:104:in `call'
	/Users/jasl/Workspaces/Ruby/Core/spec/spec_helper.rb:104:in `block in <class:Context>'
	spec/source_spec.rb:17:in `block (2 levels) in <module:Pod>'
	spec/source_spec.rb:12:in `block in <module:Pod>'
	spec/source_spec.rb:4:in `<module:Pod>'
	spec/source_spec.rb:3:in `<top (required)>'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/cli/exec.rb:63:in `load'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/cli/exec.rb:63:in `kernel_load'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/cli/exec.rb:28:in `run'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/cli.rb:497:in `exec'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/cli.rb:30:in `dispatch'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/cli.rb:24:in `start'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/bundler-2.2.3/libexec/bundle:49:in `block in <top (required)>'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/3.0.0/bundler/friendly_errors.rb:130:in `with_friendly_errors'
	/Users/jasl/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/bundler-2.2.3/libexec/bundle:37:in `<top (required)>'
	/Users/jasl/.rbenv/versions/3.0.0/bin/bundle:23:in `load'
	/Users/jasl/.rbenv/versions/3.0.0/bin/bundle:23:in `<main>'

1280 specifications (2275 requirements), 1 failures, 0 errors
```

but re-run it (and more) will success with no failures

igor-makarov:
@jasl congratulations on getting to green! Couple of comments:
* Use of `Async::Semaphore.new(4,...` causes the cold download (when the repo dir is empty) to be significantly slower. I've tried the same code without a semaphore (just using `barrier.async` instead) and it works at the normal speed. The reason for the slowdown is that HTTP2 hosts can pipeline many requests simultaneously, much more than 4.
* I'm noticing `error: Async::Task` errors printed in the test logs - can these be silenced so that they don't confuse users and CI logs?
* Is `NIO4R_PURE=true` necessary for running on M1? Because if so, that's not a good experience for users.
* Having `REQUEST_TIMEOUT = 10` seems arbitrary. Is there no default timeout in `async-http`? If it's mandatory, can we make this `ENV`-configurable? Some users will definitely want to tweak this.

jasl:
> @jasl congratulations on getting to green! Couple of comments:
> 
> * Use of `Async::Semaphore.new(4,...` causes the cold download (when the repo dir is empty) to be significantly slower. I've tried the same code without a semaphore (just using `barrier.async` instead) and it works at the normal speed. The reason for the slowdown is that HTTP2 hosts can pipeline many requests simultaneously, much more than 4.
> * I'm noticing `error: Async::Task` errors printed in the test logs - can these be silenced so that they don't confuse users and CI logs?
> * Is `NIO4R_PURE=true` necessary for running on M1? Because if so, that's not a good experience for users.
> * Having `REQUEST_TIMEOUT = 10` seems arbitrary. Is there no default timeout in `async-http`? If it's mandatory, can we make this `ENV`-configurable? Some users will definitely want to tweak this.

> Use of Async::Semaphore.new(4,... causes the cold download (when the repo dir is empty) to be significantly slower. I've tried the same code without a semaphore (just using barrier.async instead) and it works at the normal speed. The reason for the slowdown is that HTTP2 hosts can pipeline many requests simultaneously, much more than 4.

You're right, the Client has an internal pool and can be set concurrent-limit

> I'm noticing error: Async::Task errors printed in the test logs - can these be silenced so that they don't confuse users and CI logs?

In my real world test (the log I mentioned [output.log](https://github.com/CocoaPods/Core/files/5827108/output.log)) I don't see async log, especially I meet a network timeout

> Is NIO4R_PURE=true necessary for running on M1? Because if so, that's not a good experience for users.

For system ruby, yes, the root reason is nio4r c-ext can't determine the Ruby's arch... Custom Ruby don't have this problem
but I believe it's OK, we can set `NIO4R_PURE=true` in `CocoaPods/bin/pod`

igor-makarov:
@jasl I've pushed the semaphore remove.

Here's an example of log noise that I was talking about: https://github.com/CocoaPods/Core/pull/669/checks?check_run_id=1720116568#step:7:1417

Setting set `NIO4R_PURE=true` in `CocoaPods/bin/pod` can be done in a separate commit - @dnkoutso @amorde - what do you think?

jasl:
I'll create an issue to nio4r

jasl:
> Here's an example of log noise that I was talking about: https://github.com/CocoaPods/Core/pull/669/checks?check_run_id=1720116568#step:7:1417

I did some search, these expected outputs can't be supress for now

https://github.com/socketry/async/issues/63

Although I can suppress async logger, but may lost debugging info, so I suggest keep this behaivor

For users, because we catch all errors, it's safe to set logger level to `fetal`, so user won't see these noise

what do you think?

jasl:
`async-io` doesn't have global timeout setting, Socket supports timeout but hard to set it, I think I use `with_timeout` is enough

igor-makarov:
@jasl I looked at the referenced issue about errors, it seems that this may be happening because of [this line](https://github.com/socketry/async/blob/2b5d37a9e1eb2d2b1f0d500e5df406d6876c96e6/lib/async/task.rb#L272) - I'm pretty stumped as to why this library does this odd behavior.

jasl:
> @jasl I looked at the referenced issue about errors, it seems that this may be happening because of [this line](https://github.com/socketry/async/blob/2b5d37a9e1eb2d2b1f0d500e5df406d6876c96e6/lib/async/task.rb#L272) - I'm pretty stumped as to why this library does this odd behavior.

I don't fully understand either, normal errors should be catched at `rescue StandardError => error`, don't know why go next line

igor-makarov:
I'm sorry - I meant [this line](https://github.com/socketry/async/blob/2b5d37a9e1eb2d2b1f0d500e5df406d6876c96e6/lib/async/task.rb#L270) - the code calls `fail!` with `propagate=false` for `StandardError`.

igor-makarov:
@jasl I've pushed a change that does away with using `Async::Barrier` in favor of simpler code. Let me know what you think.

jasl:
> @jasl I've pushed a change that does away with using `Async::Barrier` in favor of simpler code. Let me know what you think.

It's OK in my opinion, `sync` is equal to `async and wait`, I add barrier just for clarify `here we need to wait all tasks done before go ahead`, and I believe we have unit test for the parallel downloading scenaro

jasl:
I make all CocoaPods tests pass on my M1 Mac, we must set `NIO4R_PURE=true` all the time (on M1 Mac)

Although `pod install`, `pod setup`, `pod update`, etc. works well with `nio4r` c-ext enabled (I use rbenv and Ruby 3.0)
But when I running CocoaPods tests, it crash all the time.

Log is here: [cocoapods.log](https://github.com/CocoaPods/Core/files/5839461/cocoapods.log)

So not only `ffi` has problem, all c-ext gems may meet segfault on M1...

Anyway, `nio4r` pure Ruby mode works well.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -118,33 +142,41 @@ def versions(name)
...
+            Async::Task.current.async do
+              download_file_async(pod_path_relative.join(podspec_version_path_relative).to_s)
+            rescue ::StandardError => e
+              errors << [version, e.message]
+            end
           end
-        end.compact.sort.reverse
+        end
+      ensure
+        http_client.close
```
ioquatix, line 163: You don't need to do this any more. It's implicit when the reactor goes out of scope.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -60,13 +74,23 @@ def preheat_existing_files
       files_to_update = files_definitely_to_update + deprecated_local_podspecs - ['deprecated_podspecs.txt']
       debug "CDN: #{name} Going to update #{files_to_update.count} files"
 
-      concurrent_requests_catching_errors do
+      errors = []
+
+      Sync do
```
ioquatix, line 79:
```
Sync do |task|
  # ...
  task.async do
    # ...
```

is better.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -369,133 +404,118 @@ def download_and_save_with_retries_async(partial_url, file_remote_url, etag, ret
...
+        headers << ['If-None-Match', etag]
+      end
+
+      begin
+        netrc_info = Netrc.read
+        netrc_host = URI.parse(file_remote_url).host
+        credentials = netrc_info[netrc_host]
+        if credentials
+          user, pass = credentials
+          headers << ['Authorization', "Basic #{Base64.strict_encode64("#{user}:#{pass}")}"]
```
ioquatix, line 495: You could also use https://github.com/socketry/protocol-http/blob/master/lib/protocol/http/header/authorization.rb

lib/cocoapods-core/cdn_source.rb
```diff
@@ -30,6 +30,20 @@ def initialize(repo)
       super(repo)
     end
 
+    # @return [Async::HTTP::Internet] The async HTTP client.
+    #
+    def http_client
+      @http_client ||=
+        begin
+          options = {
+            :retries => 0,
```
ioquatix, line 39: This probably isn't doing what you expect. Are you handling `Protocol::RequestFailed`?

lib/cocoapods-core/cdn_source.rb
igor-makarov: @ioquatix I've tried removing it and it has caused odd timeouts when I tried doing it in my production project.

igor-makarov:
@jasl I've pushed a change that factors out the concurrent error logic back into `concurrent_requests_catching_errors`. It uses a barrier now 🙃. Please let me know what you think.

jasl:
> @jasl I've pushed a change that factors out the concurrent error logic back into `concurrent_requests_catching_errors`. It uses a barrier now 🙃. Please let me know what you think.

LGTM

lib/cocoapods-core/cdn_source.rb
jasl: We handled `::StandardError` in https://github.com/CocoaPods/Core/pull/669/files/498c56c75481a395731909508a5f57f9db4cdc74#diff-7f437739a49b3cd41100e6671b555ea552bd12a91196f25d2cc7b1d6fd30693cR84

I think It should rescue `Protocol::RequestFailed` too

jasl: Oh I may misunderstand this, do you mean `Protocol::RequestFailed` raise on `Async::HTTP::Internet.new(**options)` ?

lib/cocoapods-core/cdn_source.rb
jasl: Switch to `Protocol::HTTP::Header::Authorization.basic(user, pass)` thanks!

lib/cocoapods-core/cdn_source.rb
jasl: Use `barrier` for both single and multiple tasks here, thanks!

lib/cocoapods-core/cdn_source.rb
jasl: Handle `Protocol::RequestFailed` when do the request

jasl:
just apply some Samuel's suggests and do cleanup useless variables

lib/cocoapods-core/cdn_source.rb
ioquatix: Probably because you are using `retries: 0`. You definitely shouldn't need this, and to take full advantage of persistent connections you should not be doing this.

lib/cocoapods-core/cdn_source.rb
ioquatix: Persistent connections can go away at any time or break for any reason. It depends on the server. HTTP/1 does not have any good way to inform the client of a graceful disconnect. So, you need to carefully handle this case.

lib/cocoapods-core/cdn_source.rb
jasl: I was suspect async-http builtin retry would break CocoaPods behavior (give user a warning when request error), just trying to remove `retries: 0`, tests still pass, so I remove `retries: 0`

igor-makarov:
@jasl I tried removing the `http_client.close` after enabling retries and it still hangs. 
I think enabling retry behavior in the client library might cause retrying on two levels (in the library and then in `CDNSource`), something that we'd want to avoid so that our server hosts don't get DDoSed. 

This is why we've got the custom retry logic with backoff. I'm not sure why changing this hasn't broken any tests. My guess is that `async-http` retries only apply to network errors and the test is testing HTTP 503 errors.

My suggestion is putting back retries=0 so that we don't accidentally break something that we can't test for.

jasl:
> @jasl I tried removing the `http_client.close` after enabling retries and it still hangs.
> I think enabling retry behavior in the client library might cause retrying on two levels (in the library and then in `CDNSource`), something that we'd want to avoid so that our server hosts don't get DDoSed.
> 
> This is why we've got the custom retry logic with backoff. I'm not sure why changing this hasn't broken any tests. My guess is that `async-http` retries only apply to network errors and the test is testing HTTP 503 errors.
> 
> My suggestion is putting back retries=0 so that we don't accidentally break something that we can't test for.

Understood, that's reasonable for me

igor-makarov:
@jasl I think I've figured out how to correct for log noise. Check it out.

jasl:
> @jasl I think I've figured out how to correct for log noise. Check it out.

LGTM, thanks!

igor-makarov:
@jasl I don't have an M1 but could you look some more into the NIO4R_PURE setting? I've looked at the code that holds it and it seems to have a special case for Windows - https://github.com/socketry/nio4r/blob/master/lib/nio.rb

Could you maybe run the specs for `async` and `async-http` repos locally on M1 without the pure version? If those fail, then maybe the better solution is to PR the `async` repo with a change to NIO engine selection.

jasl:
> https://github.com/socketry/nio4r/blob/master/lib/nio.rb

It's not fail, it's crash like ffi, I've create an issue https://github.com/socketry/nio4r/issues/259

I'd like to create a PR for nio4r, maybe I can do this on this weekend

jasl:
> @jasl I don't have an M1 but could you look some more into the NIO4R_PURE setting? I've looked at the code that holds it and it seems to have a special case for Windows - https://github.com/socketry/nio4r/blob/master/lib/nio.rb
> 
> Could you maybe run the specs for `async` and `async-http` repos locally on M1 without the pure version? If those fail, then maybe the better solution is to PR the `async` repo with a change to NIO engine selection.

Clarify more:

1. nio4r can't determine the arch because system-bundled Ruby very strange.
2. User-build Ruby can load nio4r C-ext, for `CocoaPods/Core` the tests run all green, but it will crash when running `CocoaPods` tests, but for daily usage, e.g. `pod install`, `pod update`, etc. they works fine.

igor-makarov:
@jasl

What I'd like to avoid is doing any hacks in CocoaPods to make NIO4R work on M1, as it might interfere with some other bundled gems.

Have you tried checking out `async` repo and `async-http` and running their specs? Do they also crash?

 What is `Gem::Platform.local` equal to in system Ruby on native M1? I'm thinking if it's there, we can do 
```

if ENV["NIO4R_PURE"] == "true" || (Gem.win_platform? && !defined?(JRUBY_VERSION)) || Gem::Platform.local.cpu == 'arm' # or something like that
```

jasl:
> Gem::Platform.local

> Have you tried checking out async repo and async-http and running their specs? Do they also crash?

not yet, will do later

> What is Gem::Platform.local equal to in system Ruby on native M1?

```
jasl@bogon:~/Workspaces/Ruby/Core on replace-typhoeus-to-async-http$ ruby -v
ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.arm64e-darwin20]
jasl@bogon:~/Workspaces/Ruby/Core on replace-typhoeus-to-async-http$ irb

WARNING: This version of ruby is included in macOS for compatibility with legacy software.
In future versions of macOS the ruby runtime will not be available by
default, and may require you to install an additional package.

irb(main):001:0> Gem::Platform.local
=> #<Gem::Platform:0x000000014c8c8fa0 @cpu="universal", @os="darwin", @version="20" @cpu="universal", @os="darwin", @version="20">
irb(main):002:0>
```

jasl:
Just create a PR for nio4r to workaround ARM-based macOS system Ruby https://github.com/socketry/nio4r/pull/260

I found async & aysnc-http can't pass tests on M1 Mac, but I'm using master branch, I'll try a release few hours later

igor-makarov:
@jasl I'm not sure this condition is correct - I also have `Gem::Platform.local.cpu == "universal"` on system Ruby on x86 Mac. 

Additionally, what does `Gem::Platform.local.cpu` say in custom build Ruby on your machine (mine says `x86_64`)? You said there are problems running native NIO in it as well.

jasl:
> @jasl I'm not sure this condition is correct - I also have `Gem::Platform.local.cpu == "universal"` on system Ruby on x86 Mac.
> 
> Additionally, what does `Gem::Platform.local.cpu` say in custom build Ruby on your machine (mine says `x86_64`)? You said there are problems running native NIO in it as well.

Oh I don't test on x86 Mac... so it is difficult now..

This is custom Ruby on my Mac, you see the cpu is `arm64`

```
jasl@bogon:~/Workspaces/Ruby/Core on replace-typhoeus-to-async-http$ ruby -v
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [arm64-darwin20]
jasl@bogon:~/Workspaces/Ruby/Core on replace-typhoeus-to-async-http$ irb
irb(main):001:0> Gem::Platform.local
=> #<Gem::Platform:0x000000013d9474b8 @cpu="arm64", @os="darwin", @version="20" @cpu="arm64", @os="darwin", @version="20">
```

jasl:
> @jasl I'm not sure this condition is correct - I also have `Gem::Platform.local.cpu == "universal"` on system Ruby on x86 Mac.
> 
> Additionally, what does `Gem::Platform.local.cpu` say in custom build Ruby on your machine (mine says `x86_64`)? You said there are problems running native NIO in it as well.

On the other hand, I don't know what would Samuel think.. but DON'T use system Ruby is a consensus for Ruby developers

jasl:
LOL, I don't how how to handle this now,

```
irb(main):001:0> RbConfig::CONFIG['host_cpu']
=> "x86_64"
```

macOS system Ruby running as a native ARM but said it is `x86_64`

igor-makarov:
@jasl this is frustrating, I'm sorry. What's the value of your `RUBY_PLATFORM` in both of the Rubies? 

Mine are -

Custom Ruby:
```
$ irb
irb(main):001:0> RUBY_PLATFORM
=> "x86_64-darwin19"
```

System Ruby:
```
$ /usr/bin/irb

WARNING: This version of ruby is included in macOS for compatibility with legacy software.
In future versions of macOS the ruby runtime will not be available by
default, and may require you to install an additional package.

irb(main):001:0> RUBY_PLATFORM
=> "universal.x86_64-darwin20"

**Comment by dcvz** [comment]
On an M1 machine:

Custom ruby: `arm64-darwin20`

System Ruby: `universal.arm64e-darwin20`

**Comment by igor-makarov** [comment]
Thanks @dcvz!

@jasl you did say that you have failing specs in CocoaPods even when using custom Ruby on ARM?

This means that the NIO PR should probably be something like `RUBY_PLATFORM =~ /darwin/ && RUBY_PLATFORM =~ /arm64/`. WDYT?

**Comment by dcvz** [comment]
I tried checking out the nio repo and running the specs.

Without any changes, it fails as expected. Changing the  check to: 

```ruby
if ENV["NIO4R_PURE"] == "true" || (Gem.win_platform? && !defined?(JRUBY_VERSION)) || (RUBY_PLATFORM =~ /darwin/ && RUBY_PLATFORM =~ /arm64/)
```

Gets passing results in the specs:

```
Pending: (Failures listed here are expected and do not affect your suite's status)

  1) UDPSocket behaves like an NIO selectable does not select unwritable objects
     # come up with a UDPSocket that's blocked on writing
     Failure/Error: raise TypeError, "can't convert #{io.class} into IO" unless io.is_a? IO

     TypeError:
       can't convert NilClass into IO
     Shared Example Group: "an NIO selectable" called from ./spec/nio/selectables/udp_socket_spec.rb:47
     # ./lib/nio/monitor.rb:19:in `initialize'
     # ./lib/nio/selector.rb:59:in `new'
     # ./lib/nio/selector.rb:59:in `block in register'
     # ./lib/nio/selector.rb:53:in `synchronize'
     # ./lib/nio/selector.rb:53:in `register'
     # ./spec/support/selectable_examples.rb:36:in `block (2 levels) in <top (required)>'

Finished in 8.14 seconds (files took 0.16405 seconds to load)
112 examples, 0 failures, 1 pending

Randomized with seed 64583

[Coveralls] Outside the CI environment, not sending data.
```

We can wait for @jasl to return, otherwise I can also open a PR with these changes.

jasl:
> I tried checking out the nio repo and running the specs.
> 
> Without any changes, it fails as expected. Changing the check to:
> 
> ```ruby
> if ENV["NIO4R_PURE"] == "true" || (Gem.win_platform? && !defined?(JRUBY_VERSION)) || (RUBY_PLATFORM =~ /darwin/ && RUBY_PLATFORM =~ /arm64/)
> ```
> 
> Gets passing results in the specs:
> 
> ```
> Pending: (Failures listed here are expected and do not affect your suite's status)
> 
>   1) UDPSocket behaves like an NIO selectable does not select unwritable objects
>      # come up with a UDPSocket that's blocked on writing
>      Failure/Error: raise TypeError, "can't convert #{io.class} into IO" unless io.is_a? IO
> 
>      TypeError:
>        can't convert NilClass into IO
>      Shared Example Group: "an NIO selectable" called from ./spec/nio/selectables/udp_socket_spec.rb:47
>      # ./lib/nio/monitor.rb:19:in `initialize'
>      # ./lib/nio/selector.rb:59:in `new'
>      # ./lib/nio/selector.rb:59:in `block in register'
>      # ./lib/nio/selector.rb:53:in `synchronize'
>      # ./lib/nio/selector.rb:53:in `register'
>      # ./spec/support/selectable_examples.rb:36:in `block (2 levels) in <top (required)>'
> 
> Finished in 8.14 seconds (files took 0.16405 seconds to load)
> 112 examples, 0 failures, 1 pending
> 
> Randomized with seed 64583
> 
> [Coveralls] Outside the CI environment, not sending data.
> ```
> 
> We can wait for @jasl to return, otherwise we can also open a PR with these changes.
> 
> **NOTE: tested on both system and custom ruby on M1**

Amazing! I must confess I was not do well research here.

I'll update my PR. thank you!

dcvz:
@jasl do the specs for you in this repo pass? I tried setting up a little CI for this branch and this is the outcome: https://buildkite.com/double-symmetry/cococapods-core-validation/builds/11#dc115e20-bc28-4a98-b58d-2d4863d87ccf

igor-makarov:
@dcvz Is this your own machine? Or an infra you're renting? Did you want to hook it up to CocoaPods as a proper CI check? Sorry for all the questions...

dcvz:
@igor-makarov this is a machine I own, but is dedicated to CI purposes.
I set it up because I thought it might help with this effort as you and other members don't have access to an M1.

If you invite me to the organisation I can hook it up and have it report back the status.

I'm confident I can keep it running for Cocoapods until Github Actions support an M1 :)

jasl:
> @jasl do the specs for you in this repo pass? I set up a little CI for this branch and this is the outcome: https://buildkite.com/double-symmetry/cococapods-core-validation/builds/11#dc115e20-bc28-4a98-b58d-2d4863d87ccf
> 
> _Right now, I have to trigger the CI manually since I can't hook it up to this repo._

You mean this PR?

```
1281 specifications (2326 requirements), 0 failures, 0 errors
jasl@bogon:~/Workspaces/Ruby/Core on replace-typhoeus-to-async-http$ ruby -v
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [arm64-darwin20]
```

```
1281 specifications (2326 requirements), 0 failures, 0 errors
jasl@bogon:~/Workspaces/Ruby/Core on replace-typhoeus-to-async-http$ ruby -v
ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.arm64e-darwin20]

**Comment by jasl** [comment]
> @jasl do the specs for you in this repo pass? I set up a little CI for this branch and this is the outcome: https://buildkite.com/double-symmetry/cococapods-core-validation/builds/11#dc115e20-bc28-4a98-b58d-2d4863d87ccf
> 
> _Right now, I have to trigger the CI manually since I can't hook it up to this repo._

```
/bin/bash: line 1: cd: spec/fixtures/spec-repos/test_repo: No such file or directory
```

I've met this problem, I believe it made CI fail, and it can 100% reproducible on the first run tests (after clone the repo), and only occured on the first running tests.

I'm not digging this.

**Comment by dcvz** [comment]
Yes, I meant this PR. Thanks @jasl, looks like I made a mistake in the CI script. It's all green now!
https://buildkite.com/double-symmetry/cococapods-core-validation/builds/16#bc61509b-4965-4264-9afb-cbcb51162c9e

**Comment by igor-makarov** [comment]
@dcvz First of all, I'd like to thank you for your desire to volunteer your M1 compute for the use of CocoaPods! 
I've added the webhook and tried to push a commit to test it, but nothing showed up in the pipeline page. Perhaps it's not set to build PRs? 

Anyhow, I'm a relatively junior member of CocoaPods, so I don't feel I have the authority to add new members... Access to the CocoaPods org has far-reaching security implications as far as I know. 

I was wondering if we could somehow set up a separate CocoaPods org on BuildKite and run a separate `buildkite-agent` on your infrastructure for it. This will have several advantages:
* CocoaPods has more than one repo that might benefit from M1 testing
* This will let me and other members access and control over the pipelines and other configuration 
* If your M1 infrastructure becomes unavailable, we would be able to add another with minimal disruption

**Comment by dcvz** [comment]
@igor-makarov @jasl is the only thing remaining here a new version of nio4r with the proposed changes?

**Comment by igor-makarov** [comment]
@dcvz well, it doesn't appear that Core actually fails on nio4r - @jasl this is an issue that comes up later in the CocoaPods repo, right?

**Comment by jasl** [comment]
> @igor-makarov @jasl is the only thing remaining here a new version of nio4r with the proposed changes?

I think that PR is better have, I'd propose add `NIO4R_PURE=true` on `cocoapods/bin/pod` and `bundle exec rake spec`

**Comment by jasl** [comment]
> @dcvz well, it doesn't appear that Core actually fails on nio4r - @jasl this is an issue that comes up later in the CocoaPods repo, right?

Yes, I'd propose add NIO4R_PURE=true on cocoapods/bin/pod and bundle exec rake spec

**Comment by dcvz** [comment]
Does it make sense to setup a cocoapods PR that's pointed to this branch of core? That way the whole stack can be coordinated and tested.

It would also give us a bit more time to see if nio4r merges our changes and if not, I suppose we could also have a cocoapods owned fork instead.

**Comment by igor-makarov** [comment]
@dcvz There are a few blockers to land first before this is possible - we need to finish deprecating Ruby <2.6 in both repos (PRs open) and we are gonna need to add M1 testing to CocoaPods repo as well.

**Comment by jasl** [comment]
@dcvz 

There're 2 block PRs:

- https://github.com/CocoaPods/Core/pull/671
- https://github.com/CocoaPods/CocoaPods/pull/10346

I'm finished these changes, just waiting approve & merge

**Comment by jasl** [comment]
I'm trying to do some workaround on Cocoapods side to workaround `nio4r`

**Comment by igor-makarov** [comment]
@jasl there's one additional PR that isn't even open yet - deprecating Ruby <2.6 in CocoaPods. But it'll have to wait until the corresponding Core PR is merged.

**Comment by jasl** [comment]
> @jasl there's one additional PR that isn't even open yet - deprecating Ruby <2.6 in CocoaPods. But it'll have to wait until the corresponding Core PR is merged.

Sure

**Comment by jasl** [comment]
I just added workaround for nio4r, how to trigger BuildKite CI?

**Comment by igor-makarov** [comment]
@jasl We didn't merge the pipeline definition (#672) so I can't trigger it yet. 
However, I do believe we need to discuss the workaround. If we are going to use it, we need to define it globally, otherwise it will be very fragile to maintain.

**Comment by jasl** [comment]
> @jasl We didn't merge the pipeline definition (#672) so I can't trigger it yet.
> However, I do believe we need to discuss the workaround. If we are going to use it, we need to define it globally, otherwise it will be very fragile to maintain. It is also easy to add after we merge this PR. Let's remove it for now.

OK

**Comment by igor-makarov** [comment]
@jasl In the meantime, could you (temporarily) rebase your PR on top of #672? I'd like to run the M1 tests.

**Comment by jasl** [comment]
> @jasl In the meantime, could you (temporarily) rebase your PR on top of #672? I'd like to run the M1 tests.

Done

**Comment by jasl** [comment]
@igor-makarov I rebase this PR on master accidentally, should I keep rebase on https://github.com/CocoaPods/Core/pull/672 ?

**Comment by igor-makarov** [comment]
@jasl I think we'll handle the M1 testing after this one. So let's keep it on `master`.

**Comment by jasl** [comment]
> @jasl I think we'll handle the M1 testing after this one. So let's keep it on `master`.
> 
> Also, could you please drop commits:
> 
> * `90fa064`
> * `566dfc4`
> * `0d65ae6`
> * `ca65af5`
> * `8e689d1`
> * `fac0b03`
> 
> They should all be redundant or cancel each other.

I was missing this message yesterday, I've squashed all changes into one commit

**Comment by igor-makarov** [comment]
@jasl my bad, I edited the message and assumed you'll get notified about the edit...

**Comment by jasl** [comment]
@igor-makarov 
Good news https://github.com/socketry/nio4r/pull/260 merged, when `nio4r` new version released, we don't need to workaround it

**Comment by igor-makarov** [comment]
@jasl - this was the outcome I was hoping for!
@dnkoutso - I'll create a super-issue to track all the M1-related tasks.

We really don't have much work left to do, let's get this done! 💪🏻

**Comment by igor-makarov** [comment]
@jasl I've become frustrated with trying to stem the tide of log noise from `async` in a sensible way in tests, so it's now suppressed. I've added a code comment linking to the discussion regarding the false positives that the library outputs.

**Comment by jasl** [comment]
> @jasl I've become frustrated with trying to stem the tide of log noise from `async` in a sensible way in tests, so it's now suppressed. I've added a code comment linking to the discussion regarding the false positives that the library outputs.

Yeah, I also found this, I think it's an regression of `nio4r` (affact all Ruby 2.6.x), I've discussed with Samuel in the morning, I'll look this these days

**Comment by jasl** [comment]
> @jasl I've become frustrated with trying to stem the tide of log noise from `async` in a sensible way in tests, so it's now suppressed. I've added a code comment linking to the discussion regarding the false positives that the library outputs.

could you revert the latest change to keep exposing the problem,
I'll create an issue on `nio4r`, today I'm busy, I'll back to look this tomorrow, and Samuel already known, he also will take a look when he has time.

**Comment by igor-makarov** [comment]
@jasl I've exposed the logs - so you think this is a different problem than what we had before with the "task being awaited after it completed" false positive?

**Comment by jasl** [comment]
> @jasl I've exposed the logs - so you think this is a different problem than what we had before with the "task being awaited after it completed" false positive?

I'm not think so, because only Ruby 2.6.x affact this

**Comment by igor-makarov** [comment]
@jasl Please note that now the log noise appears in a completely unrelated place and next to the `CDNSource` tests. 😬

**Comment by igor-makarov** [comment]
@jasl  I'm currently having trouble with `nio4r` native crashing on my machine when trying to actually run `pod install` in my production app repo. I suspect it to be a macOS 11.2 regression, but I'm not sure. Pure NIO works fine. https://github.com/socketry/async/issues/101

**Comment by jasl** [comment]
> @jasl I'm currently having trouble with `nio4r` native crashing on my machine when trying to actually run `pod install` in my production app repo. I suspect it to be a macOS 11.2 regression, but I'm not sure. Pure NIO works fine. [socketry/async#101](https://github.com/socketry/async/issues/101)

https://github.com/socketry/nio4r/issues/261 I've created issue and noticed Samuel, he will look this soon, it affact all Ruby 2.6.x.

I think we can pin to `nio4r 2.5.4` temporary for not blocking us for now

**Comment by igor-makarov** [comment]
@jasl - this isn't it, I'm getting a segfault in native `nio4r` extension. Downgrading to `2.5.4` doesn't help...

**Comment by jasl** [comment]
> @jasl - this isn't it, I'm getting a segfault in native `nio4r` extension. Downgrading to `2.5.4` doesn't help...

hmm It should be OK... I'll take a look

Maybe we should force add `NIO4R_PURE=true` on Pod CLI before he fixes the issue?

**Comment by igor-makarov** [comment]
@jasl Reproduced on Linux too: https://github.com/CocoaPods/Core/pull/674/checks?check_run_id=1848931012

**Comment by jasl** [comment]
@igor-makarov since https://github.com/CocoaPods/Core/pull/671 merged, I'll rebase this branch, removing duplicates commits

I have a branch of CocoaPods, should working fine on both M1 Mac and X86 Mac, I'll check the `nio4r` problem on that branch

**Comment by jasl** [comment]
Pin `nio4r 2.5.4` made the CI happy, I'll update my CocoaPods fork then

**Comment by igor-makarov** [comment]
@jasl the CI was failing because we added log noise checks - not an actual crash.
In the meantime, I'm still uncertain as to why I'm getting segfaults.

**Comment by jasl** [comment]
> @jasl the CI was failing because we added log noise checks - not an actual crash.
> In the meantime, I'm still uncertain as to why I'm getting segfaults.

I re-add your changes, CI still happy, I'm using x86 macOS 11.2, Ruby 2.6.5, `nio4r` made 2 tests failed but I'm not see segfault, what's your Mac environment?

**Comment by igor-makarov** [comment]
@jasl I've figured it out, unexpectedly. The problem was calling the `http_client.close` in the wrong place in the functions.
I've pushed a fix - the `nio4r` pin isn't necessary!

**Comment by jasl** [comment]
> @jasl I've figured it out, unexpectedly. The problem was calling the `http_client.close` in the wrong place in the functions.
> I've pushed a fix - the `nio4r` pin isn't necessary!

haha, still fully understand why, but let me revert the workaround

**Comment by igor-makarov** [comment]
@jasl checked again, the segfault is fixed by more correct `http_client.close` placement.

**Comment by jasl** [comment]
> @jasl checked again, the segfault is fixed by more correct `http_client.close` placement.

Cool!
So the `nio4r` segfault problem not block you right? it's weired I didn't see this in my Mac...

**Comment by igor-makarov** [comment]
@jasl Let's summarize:
#### Segfault Problem
Crash in `nio4r`, caused by a race condition by calling `http_client.close` outside the `Sync do` block.

The problem is resolved by placing `http_client.close` inside the `Sync do` block. No further follow-up needed.

#### Log noise failure

Not a crash, caused by my attempts to check for `async` log noise and verify that no unusual errors occur. Ultimately I don't think it makes sense to test for it, so I prefer suppressing all the `async` log output in the `CDNSource` tests. I don't think it's caused specifically by `nio4r` 2.5.5 so pinning isn't necessary. 

However, I do think that you need to drop commit `e7780fe`.

**Comment by jasl** [comment]
> e7780fe

`e7780fe` reverted, pinning also reverted.

About the noise I think that's @ioquatix 's style, I also see much error outputs on `async-` gems...

**Comment by jasl** [comment]
I think this PR is ready for review since https://github.com/CocoaPods/Core/pull/671 merged,

before this PR approved, I'll squash all changes into one single commit.

File: lib/cocoapods-core/cdn_source.rb [hunk 5]
```diff
@@ -369,104 +386,112 @@ def download_and_save_with_retries_async(partial_url, file_remote_url, etag, ret
...
+        raise Informative, "CDN: #{e.message}"
+      end
+
+      http_client.get file_remote_url, headers
     end
 
-    def exponential_backoff_async(retries)
-      sleep_async(backoff_time(retries))
+    def make_sleep(retries)
+      Async::Task.current.sleep backoff_time(retries)
```
igor-makarov, line 487: @jasl There's a problem here that wasn't caught by the tests: 

Method is named `#make_sleep` and accepts param `retries`, but all the calls are passing time in seconds, thus `#backoff_time` is happening twice, incorrectly. 

Fixes:
* rename to `#sleep_async`
* rename param to `seconds`
* remove inner call to `#backoff_time`

lib/cocoapods-core/cdn_source.rb
```diff
@@ -369,104 +386,112 @@ def download_and_save_with_retries_async(partial_url, file_remote_url, etag, ret
...
+          debug message + ", retries: #{retries - 1}"
+          make_sleep backoff_time(retries)
+          download_and_save_with_retries_async(partial_url, file_remote_url, etag, retries - 1)
         end
+      else
+        raise Informative, "CDN: #{name} URL couldn't be downloaded: #{file_remote_url} Response: #{response.status} #{body}"
+      end
+    end
+
+    def make_download(file_remote_url, etag = nil)
```
igor-makarov, line 462: @jasl Perhaps a better name is `#create_download_task`?

lib/cocoapods-core/cdn_source.rb
jasl: my bad, applied your suggests.

lib/cocoapods-core/cdn_source.rb
jasl: No problem.

spec/cdn_source_spec.rb
```diff
@@ -239,38 +195,33 @@ def typhoeus_non_http_response_future(code)
...
+          to_return(:status => 503, :headers => {}, :body => '').
+          to_return(:status => 503, :headers => {}, :body => '').
+          to_return(:status => 503, :headers => {}, :body => '').
+          to_return(:status => 200, :headers => {}, :body => 'BeaconKit/1.0.0')
+
+        WebMock.stub_request(:get, 'http://localhost:4321/Specs/2/0/9/BeaconKit/1.0.0/BeaconKit.podspec.json').
+          to_return(:status => 200, :headers => {}, :body => '')
+
+        [4, 8, 16, 32].each_with_index do |seconds, i|
+          @source.expects(:backoff_time).with(Pod::CDNSource::MAX_NUMBER_OF_RETRIES - i).returns(seconds)
```
igor-makarov, line 209: @jasl perhaps change back to the way it was?
```ruby
  [4, 8, 16, 32].each do |seconds|		
           @source.expects(:sleep_async).with(seconds)		
  end
```

spec/cdn_source_spec.rb
```diff
@@ -239,38 +195,33 @@ def typhoeus_non_http_response_future(code)
...
-        [4, 8, 16, 32].each do |seconds|
-          @source.expects(:sleep_async).with(seconds).returns(resolved_event)
+        WebMock.stub_request(:get, 'http://localhost:4321/all_pods_versions_2_0_9.txt').
+          to_return(:status => 503, :headers => {}, :body => '').
+          to_return(:status => 503, :headers => {}, :body => '').
+          to_return(:status => 503, :headers => {}, :body => '').
+          to_return(:status => 503, :headers => {}, :body => '').
+          to_return(:status => 503, :headers => {}, :body => '')
+
+        [4, 8, 16, 32].each_with_index do |seconds, i|
```
igor-makarov, line 223: @jasl same as previous one

spec/cdn_source_spec.rb
```diff
@@ -279,55 +230,48 @@ def typhoeus_non_http_response_future(code)
...
+          to_raise(SocketError).
+          to_raise(SocketError).
+          to_raise(SocketError).
+          to_return(:status => 200, :headers => {}, :body => 'BeaconKit/1.0.0')
+
+        WebMock.stub_request(:get, 'http://localhost:4321/Specs/2/0/9/BeaconKit/1.0.0/BeaconKit.podspec.json').
+          to_return(:status => 200, :headers => {}, :body => '')
+
+        [4, 8, 16, 32].each_with_index do |seconds, i|
+          @source.expects(:backoff_time).with(Pod::CDNSource::MAX_NUMBER_OF_RETRIES - i).returns(seconds)
```
igor-makarov, line 244: @jasl same as previous

spec/cdn_source_spec.rb
```diff
@@ -279,55 +230,48 @@ def typhoeus_non_http_response_future(code)
...
-          @source.expects(:sleep_async).with(seconds).returns(resolved_event)
+        WebMock.stub_request(:get, 'http://localhost:4321/all_pods_versions_2_0_9.txt').
+          to_raise(SocketError).
+          to_raise(SocketError).
+          to_raise(SocketError).
+          to_raise(SocketError).
+          to_raise(SocketError)
+
+        [4, 8, 16, 32].each_with_index do |seconds, i|
+          @source.expects(:backoff_time).with(Pod::CDNSource::MAX_NUMBER_OF_RETRIES - i).returns(seconds)
```
igor-makarov, line 259: @jasl same as previous

spec/cdn_source_spec.rb
jasl: good, simplier than current one

spec/cdn_source_spec.rb
```diff
@@ -49,39 +47,27 @@ def print_dir(tag)
...
-      end
-
       @remote_dir = fixture('mock_cdn_repo_remote')
 
       @path = fixture('spec-repos/test_cdn_repo_local')
       cleanup
       save_url('http://localhost:4321/')
 
       @source = CDNSource.new(@path)
+      @source.stubs(:sleep_async).returns(nil)
```
igor-makarov, line 57: @jasl is this stub needed?

spec/cdn_source_spec.rb
```diff
@@ -49,39 +47,27 @@ def print_dir(tag)
...
+      @save_log_level = Async.logger.level
+
+      # silence "unawaited task error" false positives to reduce log noise
+      # see https://github.com/socketry/async/issues/91 for more discussion
+      # uncomment for debug purposes
+      Async.logger.level = :fatal
     end
 
     after do
+      @source.unstub(:sleep_async)
```
igor-makarov, line 68: @jasl same as prev

spec/cdn_source_spec.rb
jasl: I'm not sure if I'm not stub `sleep_async`, it will actual do sleep, that will made tests running longer.

Let me do an experiment for that.

igor-makarov: @jasl perhaps it should be `@source.expects(:sleep_async).never`? The rest of the tests shouldn't cause sleep.

jasl: I think we should stub it, or it will do sleep

igor-makarov: @jasl perhaps it should be `@source.expects(:sleep_async).never`? The rest of the tests shouldn't cause sleep.

jasl: `raises if unexpected non-HTTP error (60.0 s)` this one

igor-makarov: @jasl I've realized this was a missing expectation, see my latest commit.

jasl: oh I forget that... thanks

ioquatix:
I think we merged most of the changes needed for this, including if you use `Sync` the error is not logged.

igor-makarov:
@jasl it looks like there's been a big change in `async` and `async-pool` which makes the `http_client.close` redundant - but this might require setting a minimum version on both of these gems. WDYT? We could just leave the code as it is, because it works.

igor-makarov:
@jasl My reason for the dilemma is because it means actually adding 2 more gems to the gemspec.

jasl:
> @jasl it looks like there's been a big change in `async` and `async-pool` which makes the `http_client.close` redundant - but this might require setting a minimum version on both of these gems. WDYT? We could just leave the code as it is, because it works.

I'm thinking setting a minimum version of `async-http` is enough:

- `gem install` should install the newest version of depenencies first, unless it already has not newest but compaitible installed. 
- `async-http' should update the requiements, if `async` and `async-pool` changes not compaitible with `async-http'

igor-makarov:
@jasl Samuel didn't bump the `async-http` requirement of the others though.

jasl:
> @jasl Samuel didn't bump the `async-http` requirement of the others though.

Don't worry for now, this PR still under reviewing, we can do your suggest at the final moment.

@ioquatix what do you think?

ioquatix:
It's fine by me, it's good to have this test case and I hope we can release 1.0 release of `async-http` this year.

lib/cocoapods-core/cdn_source.rb
```diff
@@ -369,104 +386,112 @@ def download_and_save_with_retries_async(partial_url, file_remote_url, etag, ret
...
+            debug message + ", retries: #{retries - 1}"
+            sleep_async backoff_time(retries)
+            download_and_save_with_retries_async(partial_url, file_remote_url, etag, retries - 1)
+            return
+          end
+        end
+
+      body = response.read
+      case response.status
+      when 301
```
igor-makarov, line 417: @jasl check out https://github.com/CocoaPods/Core/pull/679 - could you please add `302` here?

jasl: Added, I also rebased master and squashed commits

jasl:
long time no follow this, any news? @igor-makarov

jasl:
Just rebased master, squashed all `bundle update`

jasl:
Any news here?

segiddins (approved):

igor-makarov:
Hi @jasl - long time! 

I'm currently a bit in a dilemma - on the one hand, this PR contains much work, replacing one library with another. On the other hand - the libraries that are being replaced have already been updated for Apple Silicon - and I can personally confirm that under a correct Ruby installation, Typhoeus works natively.

There's a whole separate issue regarding the term "correct Ruby installation" as apparently the System Ruby on Apple Silicon can't work with native extensions because they get build according to RbConfig which is [incorrect](https://betterprogramming.pub/ruby-on-apple-silicon-m1-macs-fb159849b2f5).

jasl:
> Hi @jasl - long time!
> 
> I'm currently a bit in a dilemma - on the one hand, this PR contains much work, replacing one library with another. On the other hand - the libraries that are being replaced have already been updated for Apple Silicon - and I can personally confirm that under a correct Ruby installation, Typhoeus works natively.
> 
> There's a whole separate issue regarding the term "correct Ruby installation" as apparently the System Ruby on Apple Silicon can't work with native extensions because they get build according to RbConfig which is [incorrect](https://betterprogramming.pub/ruby-on-apple-silicon-m1-macs-fb159849b2f5).

I believe `nio4r` can handle both `x86` and `arm64` well, this can easily test

igor-makarov:
@jasl this isn't what worries me - I have tested `nio4r` on both. What worries me is that we're talking about shipping a HUGE change in a core module - while we actually don't have to.

@segiddins what do you think?

ioquatix:
If it’s any help we found problems with Typhoeus and it seems largely unmaintained.

I haven’t looked closely at this PR for a while but I would say you could adopt it if you build an abstract interface for doing the requests and then support both. That way can compare performance too which should be interesting.

Such an approach might make migration easier.

igor-makarov:
@ioquatix I'm less worried about the performance impact - there are well-defined places where concurrency is utilized and it's easy to see that it works correctly. I'm more worried about the fact that it's a huge change in a core component and the impact of even a small bug can be catastrophic.

Having 2 backends will increase the amount of work even more that has been invested right now and increase the test surface even more.

The Typhoeus implementation works for the project's needs and it's not a problem to load it on ASi architecture anymore.

amorde:
Having `ffi` has a dependency has introduced friction in more ways than just the introduction of Apple Silicon, so I'd love to remove it. That said I'm not personally familiar with the differences in these two libraries and how much risk there is. I'd suggest releasing it in a few betas to get a sense of stability, and then ship it, as long as we are available to triage any issues that come up

MaxDesiatov:
Is there anything blocking this PR? Is anything else needed other than resolving conflicts?

amorde:
It's mostly a matter of our ability to maintain this going forward + push out potential bug fix releases if we find issues after releasing it.

There's been more issues with `ffi` recently with the release of macOS Monterey and Xcode 13. I think it is still worthwhile to pursue removing our dependency on `ffi`

jasl:
I have business traval these days, and will back home next week, I can help to rebase to master branch to fix conflicts

igor-makarov:
@amorde can you point me to the new `ffi` issues? Haven't heard about any of it unfortunately.

amorde:
We had some internal issues with our CI machines that seem to be related to this https://twitter.com/churowa/status/1455971248555511808

igor-makarov:
I've gone over this PR one more time and there are several issues still related to using `async-http`:

* There's no easy way to use an outgoing proxy like there is with `libcurl`/`typhoeus`. I've researched this and it seems that the `async-http` library added minimal support for this, that cannot work with HTTPS at all.
* There's no library support for `netrc` - the hand-rolled implementation may break and we may be required to fix it.

In addition to these gaps, the change is **huge** and we definitely don't have sufficient tests in place to verify everything.

Much time has passed since the M1 Macs first emerged and there ***exist*** versions of `ffi` and `typhoeus` that are working correctly on the architecture, even if some CI vendors are having trouble deploying them.

It is my opinion that because of all the reasons I've listed above, we should not commit to making this change. 

Big thanks to @jasl for all the hard work done on this, and my sincere apology for not wanting to merge it. 🙏🏻

jasl:
> I've gone over this PR one more time and there are several issues still related to using `async-http`:
> 
> * There's no easy way to use an outgoing proxy like there is with `libcurl`/`typhoeus`. I've researched this and it seems that the `async-http` library added minimal support for this, that cannot work with HTTPS at all.
> * There's no library support for `netrc` - the hand-rolled implementation may break and we may be required to fix it.
> 
> In addition to these gaps, the change is **huge** and we definitely don't have sufficient tests in place to verify everything.
> 
> Much time has passed since the M1 Macs first emerged and there _**exist**_ versions of `ffi` and `typhoeus` that are working correctly on the architecture, even if some CI vendors are having trouble deploying them.
> 
> It is my opinion that because of all the reasons I've listed above, we should not commit to making this change.
> 
> Big thanks to @jasl for all the hard work done on this, and my sincere apology for not wanting to merge it. 🙏🏻

it's OK, the best is FFI can working good on M1.

For your opinions about `async`, I should forward to @ioquatix , maybe can be improve in the future, personally proxy support is also important to me.

amorde:
Thanks for taking another look @igor-makarov. Do the same points apply to your original PR that uses `net-http2`? https://github.com/CocoaPods/Core/pull/639

igor-makarov:
@amorde To be honest it was a draft and I don't think I was close to finishing it 🤷🏻‍♂️

amorde:
Ok makes sense. I still do wish we didn't depend on ffi as it tends to be a common source of issues but I understand the reason we aren't there yet.

dnkoutso:
Should we close those PRs?

jasl:
> Ok makes sense. I still do wish we didn't depend on ffi as it tends to be a common source of issues but I understand the reason we aren't there yet.

As a Rubyist, I wish `ffi` gem can running on M1 smoothly, many gems depend on it.

But for a CLI tool, less C-ext dependencies, less trouble to end-users

igor-makarov:
Yes, I would have also preferred a Ruby-native implementation, but it's very hard to compete with `libcurl` in terms of feature completeness. It's truly a Swedish army knife.

ioquatix:
> There's no easy way to use an outgoing proxy like there is with libcurl/typhoeus. I've researched this and it seems that the async-http library added minimal support for this, that cannot work with HTTPS at all.

async-http has full support for HTTP/1 and HTTP/2 proxies including TLS.

> There's no library support for netrc - the hand-rolled implementation may break and we may be required to fix it.

Is this what you are referring to? <https://everything.curl.dev/usingcurl/netrc>

It seems awfully insecure. I don't think we'd want to support this by default.

igor-makarov:
@ioquatix I've seen that the library does offer proxy support. However, I've spent most of yesterday trying to allow to users to enable it as a simple `http_proxy` env var, and I've not been successful.  
Specifically, I've gotten to the point where the outgoing connection was being established, and then failing due to TLS validation. I wasn't able to figure out how to disable certificate validation for my `localhost` proxy.

As for `netrc` - a lot of the users are using it, and depend on it to authenticate with their private spec repos and Artifactories. It's a def facto standard for basic auth and CocoaPods needs to support it.

PR #670, opened 2021-01-12
Use Ruby bundled json gem instead of forked

jasl:
Same as https://github.com/CocoaPods/CocoaPods/pull/10347

It looks like we no longer need that fork, trying to remove it


jasl:
CI Green

segiddins (approved):

amorde (approved):

PR #681, opened 2021-02-27
Normalize spec repo name generation

igor-makarov:
This PR tries to simplify the way that a spec repo URL is converted to a repo name. 

The main cases covered:

* GitHub URL: `https://github.com/Orgname/Specs.git` => `orgname`
* Non-GH URL: `https://sourceforge.org/Orgname/Specs.git` => `sourceforge-orgname`
* SCP-style URL: `git@github.com/Orgname/Specs` => `orgname`
* File URL: `file:///Users/username/specs` => `users-username`
* Other odd URLs, such as `ssh://user@companyalias/pod-specs`
* If the repo isn't named `Specs`: `https://github.com/Username/banana.git` => `username-banana`

### DIFF
--- a/lib/cocoapods-core/source/manager.rb
+++ b/lib/cocoapods-core/source/manager.rb
@@ -434,7 +434,7 @@ module Pod
           end
 
           path = path.gsub(/.git$/, '').gsub(%r{^/}, '').split('/')
-          path.delete('specs')
+          path.pop if path.last == 'specs'
 
           (base + path).join('-')
         end
--- a/lib/cocoapods-core/source/manager.rb
+++ b/lib/cocoapods-core/source/manager.rb
@@ -1,3 +1,5 @@
+require 'public_suffix'
+
 module Pod
   class Source
     class Manager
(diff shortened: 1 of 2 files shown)


lib/cocoapods-core/source/manager.rb
```diff
@@ -424,28 +426,48 @@ def canonic_url(url)
...
+            base = [domain&.sld || host]
+            base = [] if base == %w(github)
           else
-            base = ''
+            base = []
           end
 
-          base + path.gsub(/.git$/, '').gsub(%r{^/}, '').split('/').join('-')
+          path = path.gsub(/.git$/, '').gsub(%r{^/}, '').split('/')
+          path.delete('specs')
```
segiddins, line 437: should this be `path.pop if path.last == 'specs'`, to handle a `specs` segment in the middle of the path? can we add specs for that case?

igor-makarov: @segiddins gotcha - added!

igor-makarov:
@segiddins I've also removed `infinity` in favor of an open range `(1..)` which is available in Ruby 2.6.

segiddins (approved):

dnkoutso (approved):

PR #733, opened 2023-01-09
Bump activesupport requirement to < 8

jonabc:
:wave: I maintain a tool called [licensed](https://github.com/github/licensed) that enumerates OSS license and other metadata for dependencies used by a project.  We recently added cocoapods and received a [report](https://github.com/github/licensed/issues/609) that the `activesupport < 7` requirement from this gem was causing a user to be unable to upgrade to the latest version of licensed.

Would you consider updating the activesupport requirement to `<8` to support users that want to use this gem in conjunction with rails 7?  I _think_ this change should be ok?  It looks like the gem was only used to add indifferent access support for hashes so the impact would be minimal, and I ran the test suite with `bundle exec rake spec` and everything passed.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     cocoapods-core (1.11.3)
-      activesupport (>= 5.0, < 7)
+      activesupport (>= 5.0, < 8)
       addressable (~> 2.8)
       algoliasearch (~> 1.0)
       concurrent-ruby (~> 1.1)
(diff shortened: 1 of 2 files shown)


jonabc:
@dnkoutso apologies for the direct ping, it looks like you've merged a few of the recent changes in this repo.  Is there any chance you could take a look at this change?

If it looks alright and is ok to merge, what is the release process for this gem like and how often are releases put out?

dnkoutso:
I think thats OK? We should do it for the main cocoapods repo as well I believe.

PR #736, opened 2023-02-27
Give the pathname for failed json podspecs in the error

orta:
Definitely in the top 5 of issues we get which are pretty much unactionable are ones where JSON isn't parsed, so now we give the path to the file in the error message so its more likely people will try and solve themselves instead of making an issue

### DIFF
--- a/lib/cocoapods-core/specification/json.rb
+++ b/lib/cocoapods-core/specification/json.rb
@@ -56,10 +56,17 @@ module Pod
     #
     # @return [Specification] the specification
     #
-    def self.from_json(json)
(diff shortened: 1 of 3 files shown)


lib/cocoapods-core/specification/json.rb
```diff
@@ -56,10 +56,17 @@ def to_hash
     #
     # @return [Specification] the specification
     #
-    def self.from_json(json)
+    def self.from_json(json, path="")
```
orta, line 59: made it optional for backwards compat

PR #741, opened 2023-03-28
[YAMLHelper] Handle YAML timestamps

jaysoffian:
With Ruby 3.1, psyche uses safe_load by default. This prevents YAML timestamps from loading as they are not serialized as strings.

- Add the YAML timestamp regex to the list of RESOLVED_TAGS so that tags which look like timestamps are serialized as strings.
- Switch load_string to use YAML.safe_load so that the behavior is consistent between Ruby (psyche) versions.
- Add Date, Time, and Symbol as permitted_classes so that previously serialized  Podfile.lock files can be loaded.

References:
- https://yaml.org/type/timestamp.html
- https://github.com/ruby/psych/commit/176494297f3f124467a6e3f1c9e6400ee742d663
- https://stackoverflow.com/questions/71332602/upgrading-to-ruby-3-1-causes-psychdisallowedclass-exception-when-using-yaml-lo

### DIFF
--- a/lib/cocoapods-core/yaml_helper.rb
+++ b/lib/cocoapods-core/yaml_helper.rb
@@ -54,7 +54,7 @@ module Pod
       # @return [Hash, Array] the Ruby YAML representaton
       #
       def load_string(yaml_string, file_path = nil)
-        YAML.load(yaml_string)
+        YAML.safe_load(yaml_string, :permitted_classes => [Date, Time, Symbol])
       rescue
         if yaml_has_merge_error?(yaml_string)
           raise Informative, yaml_merge_conflict_msg(yaml_string, file_path)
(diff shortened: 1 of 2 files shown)


jaysoffian:
FWIW, this took longer to track down than it should have due to the original yaml error being
swallowed by `load_string` and it raising `yaml_parsing_error_msg` instead. It's fine to add
additional information to an exception but swallowing the underlying exception makes 
debugging harder.

dnkoutso:
Thanks! is it possible to add a test?

jaysoffian:
> Thanks! is it possible to add a test?

I'll see what I can do. I don't have a lot more time to spend on this and I'm not much of a Ruby coder. I gather https://github.com/CocoaPods/Core/blob/master/spec/yaml_helper_spec.rb is the right place to add the test?

FWIW, I tested this locally with a `Podfile.lock` that was failing for us because we had a tag that looked like `YYYY-MM-DD` under homebrew-installed CocoaPods (Ruby 3.2). Once I got it working locally I made the edit directly in the GitHub web UI by copy/pasting from my editor. Ugh... I know.

jaysoffian:
@dnkoutso test added.

jaysoffian:
Tap tap tap... is this thing on?

dnkoutso:
I approved to run CI this weekend, seems like some tests are failing.

jaysoffian:
Fixed tests (I had the quoting wrong) and rebased.

dnkoutso:
@jaysoffian some tests still failing. You can run any `_spec.rb` file using `bundle exec bacon <path/to/file>` to verify the tests pass.

segiddins (approved):

jaysoffian:
@dnkoutso Thanks. Tested locally. There were two issues.

1. The regexp on https://yaml.org/type/timestamp.html doesn't match all of the examples. In particular it did not match `2001-12-14 21:59:43.10 -5` due to the space before the timezone. Adding an extra pair of parens to the last line of the regexp fixes that.
2. I had to adjust the expectation for the `handles objects of unknown classes` test which now raises an exception instead of returning a meaningless string.

jaysoffian:
Done. Sorry about all the back-and-forth on this. I should've taken care of these issues in the first place. You can probably tell this is my first CocoaPods/ruby contribution.

dnkoutso:
It's perfectly alright!

PR #745, opened 2023-06-26
Add `visionos` as a new platform

gabrieldonadel:
This PR introduces `visionOS` as a supported platform in Core. This primarily involves making the DSL helpers accessible for this platform.  This lays the groundwork for supporting visionOS as a new platform in CocoaPods.

Related to https://github.com/CocoaPods/CocoaPods/issues/11961, https://github.com/CocoaPods/Xcodeproj/pull/913 and https://github.com/CocoaPods/CocoaPods/pull/11965

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Add `visionOS` as a new platform.
+  [Gabriel Donadel](https://github.com/gabrieldonadel)
+  [Core#745](https://github.com/CocoaPods/Core/pull/745)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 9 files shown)


dnkoutso:
We might need to bump the min version of Xcodeproj here required going forward as older versions wont work with visionOS.

gabrieldonadel:
> We might need to bump the min version of Xcodeproj here required going forward as older versions wont work with visionOS.

Tests should be fixed now
image

Where should I bump the min version of Xcodeproj? Is that set inside `CocoaPods/Core` or `CocoaPods/Xcodeproj`?

dnkoutso:
Once I make a release of Xcodeproj we should update this https://github.com/CocoaPods/CocoaPods/blob/master/cocoapods.gemspec#L40

NachoSoto:
I don't think this is fully ready: https://github.com/CocoaPods/CocoaPods/pull/11965#issuecomment-1743878949

PR #754, opened 2023-10-11
Fix semantic version comparison for mixed numeric and string prerelease segments

### LINKED ISSUE #752: Version Comparison Issue with Mixed Numeric and String Prerelease Segments
Per the [Semantic Versioning Specification](https://semver.org/), prerelease versions `1.0.0-beta.3rd.1` and `1.0.0-beta.ios17.1` are different, the first expected to be less than the second one according to spec [11.4.3](https://semver.org/#spec-item-11). However, CocoaPods currently treats these two versions as equal, leading to the installation of `1.0.0-beta.ios17.1` over `1.0.0-beta.3rd.1` when `pod foo, ‘1.0.0-beta.3rd.1’` is used in the Podfile.

The issue lies in the `compare_segments` method of `Pod::Version`. The method fails to handle a case where the prerelease segments contain both `String` and `Numeric` elements, leading to an incorrect equality comparison when the types of elements at the same index differ. 

Here's an example illustrating this issue:

```ruby
Version.new('1.0.0-beta.3rd.1').prerelease_segments
# [beta, 3, rd, 1]

Version.new('1.0.0-beta.ios17.1').prerelease_segments
# [beta, ios, 17, 1]

# compare steps
'beta' == 'beta' # loop:0, true, next
3 <=> 'ios' # loop:1, nil, next
'rd' <=> '17' # loop:2, nil, next
1 == 1 # loop:3, true, next, exit
```

I've addressed this by adding a check in the `lhs <=> rhs` comparison to handle cases where it returns `nil`, determining the types of the two elements and returning the correct result per spec [11.4.3](https://semver.org/#spec-item-11):

`lib/cocoapods-core/version.rb`
image

Fix on #754

sagiwei:
Fixes #752

### DIFF
--- a/lib/cocoapods-core/version.rb
+++ b/lib/cocoapods-core/version.rb
@@ -225,6 +225,9 @@ module Pod
 
           if comparison = lhs <=> rhs
             return comparison
+          else
+            return -1 if lhs.is_a?(Numeric) && rhs.is_a?(String)
+            return  1 if rhs.is_a?(Numeric) && lhs.is_a?(String)
           end
         end
       end
--- a/spec/version_spec.rb
+++ b/spec/version_spec.rb
@@ -145,6 +145,9 @@ module Pod
 
         # Example from CocoaPods/CocoaPods#5718
         Version.new('1.0-beta.8').should.be < Version.new('1.0-beta.8a')
+
+        # Example from section 11.4.3 on semver.org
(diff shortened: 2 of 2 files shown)


sagiwei:
@dnkoutso Can you check this please?

amorde (approved):
Based on the spec this seems correct to me. Thanks!

